PR #19, opened 2021-09-08
use SuiteSparseGraphBLAS's sparsematrix implementation (GBMatrix)

CarloLucibello:
This uses  [SuiteSparseGraphBLAS.GBMatrix](https://graphblas.juliasparse.org/dev/) as a replacement for SparseArryas.SparseMatrixCSC. 
It looks like many features are missing in  GBMatrix though and a ton of tests are not passing.

### DIFF
--- a/src/GNNGraphs/convert.jl
+++ b/src/GNNGraphs/convert.jl
@@ -124,7 +124,7 @@ function to_sparse(A::ADJMAT_T, T::DataType=eltype(A); dir=:out, num_nodes=nothi
         A = T.(A)
     end
     if !(A isa AbstractSparseMatrix)
-        A = sparse(A)
+        A = GBMatrix(sparse(A))
     end 
     return A, num_nodes, num_edges
 end
@@ -140,7 +140,7 @@ function to_sparse(coo::COO_T, T::DataType=Int; dir=:out, num_nodes=nothing)
     num_nodes = isnothing(num_nodes) ? max(maximum(s), maximum(t)) : num_nodes 
     A = sparse(s, t, eweight, num_nodes, num_nodes)
     num_edges = length(s)
-    return A, num_nodes, num_edges
+    return GBMatrix(A), num_nodes, num_edges
 end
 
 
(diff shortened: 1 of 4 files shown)


Wimmerer:
Can you assign this to me @CarloLucibello?

CarloLucibello:
Sure, thanks! I rebased on master and I guess you should be able to push to this branch now

CarloLucibello:
Ideally, I'd like any sparse matrix backend to act as a drop-in replacement of SparseMatrixCSC, so that it should be possible to easily switch between backends and compare performance. We could even consider introducing a new `graph_type = :sparseblas`.

Then one can look into specializing `propagate` on specific operations and graph types, as done here
https://github.com/CarloLucibello/GraphNeuralNetworks.jl/blob/5d53d0585c319e5f7dc574e2e393e52e8473146a/src/msgpass.jl#L157 

Unfortunately, I think decent gpu support won't be available for a while for any sparse matrix type (although things are moving a bit in CUDA.jl)

src/GNNGraphs/convert.jl
```diff
@@ -124,7 +124,7 @@ function to_sparse(A::ADJMAT_T, T::DataType=eltype(A); dir=:out, num_nodes=nothi
         A = T.(A)
     end
     if !(A isa AbstractSparseMatrix)
-        A = sparse(A)
+        A = GBMatrix(sparse(A))
```
CarloLucibello, line 127:
```suggestion
        A = GBMatrix(A)
```
?

Wimmerer:
I'll have a chance to look at this later in the week. I'll check with the upstream author about the status of the GPU portion, but it'll take a bit for sure.

CarloLucibello:
> I'll check with the upstream author about the status of the GPU portion, but it'll take a bit for sure.

it doesn't have to come all together. Actually I would keep the two things separate: first PR we integrate cpu GBMatrix support, then we see what we can do about gpus

CarloLucibello:
SuiteSparse's license seems complicated 
https://github.com/DrTimothyAldenDavis/SuiteSparse/blob/master/LICENSE.txt
Is this compatible with our MIT license?

Wimmerer:
We're only using the GraphBLAS part which is compatible (and pretty much separate) IIRC it's Apache 2. SuiteSparse's many other parts are mostly GPL, but we're safe, they are compiled separately and not linked.

I did accidentally include SuiteSparse.jl as a dependency in my most recent commit to #master so this is a good remind :).

CarloLucibello:
closing in favor of #222

Issue #29, opened 2021-09-16
define a `message_and_aggregate` method


CarloLucibello:
In order to avoid feature allocations on each edge, we should provide define a `message_and_aggregate` function
fusing together the `compute_message` and `aggregate_neighbors` functions. 

The operation to be fused together are
```julia
s, t = edge_index(g)
xi = gather(x, t)
xj = gather(x, s)
m = compute_message(l, xi, xj, e)
scatter(aggr, m, t)
```

CarloLucibello:
I think this can be considered closed by #47

PR #39, opened 2021-09-19
fuse message and aggregate

CarloLucibello:
Attempt at #29 
Higher order AD will be needed for differentiation with generic function/operators


CarloLucibello:
closing as outdated

PR #46, opened 2021-09-22
implement ResGatedGraphConv and support Parallel in GNNChain

CarloLucibello:
cc @r3tex


r3tex:
FYI - you included double parentheses around the link to the paper so it doesn't redirect correctly.

PR #109, opened 2022-01-19
Typecast GNNGraph.num_nodes to Int

abieler:
Fixes type of `GNNGraph.num_nodes` to `Int`.
Was an issue for the following example:

```julia
using Graphs
using GraphNeuralNetworks

GNNGraph(SimpleGraph{Int32}(6))
```

which would lead to `num_nodes` being of type `Int32`. (and `num_graphs`, `num_edges` were still `Int`).
Added a short comment because you might want to typecast `s`, `t` vectors to `Int`, which would also solve the issue (i guess)

Let me know what you think, or feel free to hijack/adapt the PR and merge :)

Cheers
Andre

### DIFF
--- a/src/GNNGraphs/gnngraph.jl
+++ b/src/GNNGraphs/gnngraph.jl
@@ -168,7 +168,8 @@ function GNNGraph(g::AbstractGraph; kws...)
         # add reverse edges since GNNGraph is directed
         s, t = [s; t], [t; s]    
     end
-    GNNGraph((s, t); num_nodes=Graphs.nv(g), kws...)
+    num_nodes::Int = Graphs.nv(g)
+    GNNGraph((s, t); num_nodes=num_nodes, kws...)
 end
 
 
--- a/src/GNNGraphs/convert.jl
+++ b/src/GNNGraphs/convert.jl
@@ -2,7 +2,7 @@
 
 function to_coo(coo::COO_T; dir=:out, num_nodes=nothing, weighted=true)
     s, t, val = coo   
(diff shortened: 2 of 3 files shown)


src/GNNGraphs/gnngraph.jl
```diff
@@ -162,6 +162,7 @@ GNNGraph((s, t)::NTuple{2}; kws...) = GNNGraph((s, t, nothing); kws...)
 # GNNGraph(g::AbstractGraph; kws...) = GNNGraph(adjacency_matrix(g, dir=:out); kws...)
 
 function GNNGraph(g::AbstractGraph; kws...)
+    # or maybe force eltype(s|t) == Int here?
```
CarloLucibello, line 165:
```suggestion
```

CarloLucibello:
Seems a reasonable solution, and I think it is fine to leave s,t as Int32 since it looks like that would be intended behavior.
Could you add a dummy test like
```julia
g = GNNGraph(SimpleGraph{Int32}(6), graph_type=GRAPH_T)
@test g.num_nodes == 6
```
somewhere [here](https://github.com/CarloLucibello/GraphNeuralNetworks.jl/blob/master/test/GNNGraphs/gnngraph.jl)
just to check we don't error out anymore?

CarloLucibello:
Looks like a similar fix should go also in 
```julia
to_dense(coo::COO_T,...)
to_sparse(coo::COO_T,...)
```
Sorry for that, supporting multiple backends is very annoying

abieler:
yeah just saw that. will add the fix once i am off work later today.

PR #121, opened 2022-01-28
implement hash for GNNGraph

CarloLucibello:
Make sure that `g` and `deepcopy(g)` are hashed the same. This is relevant when graphs are used as dictionary keys (e.g. when used as states in the implementation of the MCTS in AlphaZero.jl).

cc @bhalonen

### DIFF
--- a/test/layers/conv.jl
+++ b/test/layers/conv.jl
@@ -98,7 +98,7 @@
         for heads in (1, 2), concat in (true, false)
             l = GATConv(in_channel => out_channel; heads, concat)
             for g in test_graphs
-                test_layer(l, g, rtol=1e-4,
+                test_layer(l, g, rtol=1e-3,
                     outsize=(concat ? heads*out_channel : out_channel, g.num_nodes))
             end
         end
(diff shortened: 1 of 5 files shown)


test/layers/conv.jl
```diff
@@ -98,7 +98,7 @@
         for heads in (1, 2), concat in (true, false)
             l = GATConv(in_channel => out_channel; heads, concat)
             for g in test_graphs
-                test_layer(l, g, rtol=1e-3, atol=1e-3,
+                test_layer(l, g, rtol=1e-4,
```
CarloLucibello, line 101:
```suggestion
                test_layer(l, g, rtol=1e-3,
```

test/layers/conv.jl
```diff
@@ -114,7 +114,7 @@
         for heads in (1, 2), concat in (true, false)
             l = GATv2Conv(in_channel => out_channel; heads, concat)
             for g in test_graphs
-                test_layer(l, g, rtol=1e-3, atol=1e-3,
+                test_layer(l, g, rtol=1e-4,
```
CarloLucibello, line 117:
```suggestion
                test_layer(l, g, rtol=1e-3,
```

PR #142, opened 2022-03-05
drop LearnBase in favor of MLUtils

CarloLucibello:
Maybe we want to wait for Flux 0.13 (coming also with MLUtils) before merging this. 

In fact with Flux 0.12 we have
```julia
julia> using GraphNeuralNetworks
[ Info: Precompiling GraphNeuralNetworks [cffab07f-9bc2-4db1-8861-388f63bf7694]
WARNING: Method definition rpad(AbstractArray{T, 1} where T, Integer, Any) in module Flux at /home/carlo/.julia/packages/Flux/qAdFM/src/utils.jl:610 overwritten in module MLUtils at /home/carlo/.julia/packages/MLUtils/OojOS/src/utils.jl:384.
  ** incremental compilation may be fatally broken for this module **
```


CarloLucibello:
done in #155

PR #147, opened 2022-03-21
Added GMMConv

melioristic:
Added GMMConv from the paper: Geometric deep learning on graphs and manifolds using mixture model CNNs


src/layers/conv.jl
```diff
@@ -1065,4 +1065,117 @@ function (l::MEGNetConv)(g::GNNGraph, x::AbstractMatrix, e::AbstractMatrix)
...
+
+Performs the operation
+```math
+\mathbf{x}_i' = \frac{1}{|N(i)|} \sum_{j\in N(i)}\frac{1}{K}\sum_{k=1}^k\mathbf{w}_k(u_{i,j})\odot \mathbf{\theta}_k \mathbf{x}_j
+```
+
+where
+```math
+\mathbf{w_{k}} = \exp(\frac{-1}{2}(u-\mu_k)^T \Sigma_k^{-1}(u-\mu
+_k))
```
CarloLucibello, line 1081:
```suggestion
\mathbf{w_{k}} = \exp(\frac{-1}{2}(u-\mu_k)^T \Sigma_k^{-1}(u-\mu_k))
```
Also, \mu and \Sigma need to be specified

src/layers/conv.jl
```diff
@@ -1065,4 +1065,117 @@ function (l::MEGNetConv)(g::GNNGraph, x::AbstractMatrix, e::AbstractMatrix)
...
 end
 
+@doc raw"""
+    GMMConv(in => out, n_kernel, u_dim, σ=identity; [init, bias])
+
+Graph mixture model convolution layer from the paper [Geometric deep learning on graphs and manifolds using mixture model CNNs](https://arxiv.org/abs/1611.08402)
+
+Performs the operation
+```math
+\mathbf{x}_i' = \frac{1}{|N(i)|} \sum_{j\in N(i)}\frac{1}{K}\sum_{k=1}^k\mathbf{w}_k(u_{i,j})\odot \mathbf{\theta}_k \mathbf{x}_j
```
CarloLucibello, line 1075:
```suggestion
\mathbf{x}_i' = \frac{1}{|N(i)|} \sum_{j\in N(i)}\frac{1}{K}\sum_{k=1}^k w_k(u_{j\to i}) \odot \mathbf{\theta}_k \mathbf{x}_j
```

src/layers/conv.jl
CarloLucibello: no actually is not a scalar

CarloLucibello: Ok, I edited the suggestion so that now it should be mathematically clearer what's going on

src/layers/conv.jl
```diff
@@ -1065,4 +1065,117 @@ function (l::MEGNetConv)(g::GNNGraph, x::AbstractMatrix, e::AbstractMatrix)
...
+                bias::Bool=true)
+    in, out = ch
+    mu = init(n_kernel, u_dim)
+    sigma_inv = init(n_kernel, u_dim)
+    b = bias ? Flux.create_bias(ones(out), true) : false
+    dense_x = Dense(in, out*n_kernel, bias=false)
+    GMMConv(mu, sigma_inv, b, σ, ch, n_kernel, u_dim, dense_x)
+end
 
+function (l::GMMConv)(g::GNNGraph, x::AbstractMatrix, u::AbstractMatrix)
```
CarloLucibello, line 1147: let's rename `u` to `e` since we usually call by that the edge features

src/layers/conv.jl
```diff
@@ -1065,4 +1065,117 @@ function (l::MEGNetConv)(g::GNNGraph, x::AbstractMatrix, e::AbstractMatrix)
...
+
+    if g.graph[3]!==nothing
+        @warn g.graph[3] !== nothing "Edge weights $(g.graph[3]) provided, will not be used for computation"
+    end
```
CarloLucibello, line 1151:
```suggestion
```
we can remove this, weight are not used unless explicitly stated, no need to warn

src/layers/conv.jl
```diff
@@ -1065,4 +1065,117 @@ function (l::MEGNetConv)(g::GNNGraph, x::AbstractMatrix, e::AbstractMatrix)
...
+    
+    @assert (l.u_dim == size(u)[1] && g.num_edges == size(u)[2]) "Pseudo-cordinate dim $(size(u)) does not match (u_dim=$(u_dim),num_edge=$(g.num_edges))"
+
+    num_edges = g.num_edges
+    d = degree(g, dir=:in)
+    u = reshape(u,(num_edges, 1, l.u_dim))
```
CarloLucibello, line 1157:
```suggestion
    u = reshape(u, (num_edges, 1, l.u_dim))
```

CarloLucibello: The reshape is wrong actually. While in python edge features are of size `num_edges x dim`, for us it is `dim x num_edges`. So you should reshape 
```julia
u = reshape(u, (l.u_dim, 1, l.num_edges))
mu = reshape(mu, (l.u_dim, l.n_kernel, 1))
```

src/layers/conv.jl
```diff
@@ -1065,4 +1065,117 @@ function (l::MEGNetConv)(g::GNNGraph, x::AbstractMatrix, e::AbstractMatrix)
...
+    mu = reshape(l.mu, (1, l.n_kernel, l.u_dim))
+
+    w = -0.5*(u.-mu).^2
+    w = w .* ((reshape(l.sigma_inv, (1, l.n_kernel, l.u_dim)).^2) )
+    w = exp.(sum(w, dims = 3 )) # n_edges, n_kernel,  1
+    w = permutedims(w, [3,2,1])
```
CarloLucibello, line 1163: ideally if reshape is done properly there should be no need for permutation

src/layers/conv.jl
```diff
@@ -1065,4 +1065,117 @@ function (l::MEGNetConv)(g::GNNGraph, x::AbstractMatrix, e::AbstractMatrix)
...
+    
+    xj = reshape(l.dense_x(x), (l.ch[2],l.n_kernel,:))
+    
+    x = propagate(e_mul_xj, g, +, xj=xj, e=w)
+    x = dropdims(mean(x, dims=2), dims=2)
+    x = 1/d.*x
```
CarloLucibello, line 1169:
```suggestion
    x = 1 / d .* x
```

CarloLucibello (changes requested):

CarloLucibello:
Tests are missing

src/layers/conv.jl
melioristic: Thanks!

src/layers/conv.jl
melioristic: Will do that in the next commit

melioristic:
Added the changes.
- Added the learnable parameter in the doc
- Changed the name from u to e in view of the package convention 
- As a convention, [last dimension as num_edges.](https://github.com/CarloLucibello/GraphNeuralNetworks.jl/pull/147#discussion_r831118362)

Still need to add the test.

src/layers/conv.jl
```diff
@@ -1065,4 +1065,103 @@ function (l::MEGNetConv)(g::GNNGraph, x::AbstractMatrix, e::AbstractMatrix)
...
+g = GNNGraph(s,t)
+in_feature, out_feature, n_k, e_dim = 4, 7, 8, 10
+x = randn(in_feature, g.num_nodes)
+u = randn(e_dim, g.num_edges)
+
+# create layer
+l = GMMConv(in_feature=>out_feature, n_k, e_dim)
+
+# forward pass
+l(g, x, u)
```
CarloLucibello, line 1107:
```suggestion
e = randn(e_dim, g.num_edges)

# create layer
l = GMMConv(in_feature=>out_feature, n_k, e_dim)

# forward pass
l(g, x, e)
```

src/layers/conv.jl
```diff
@@ -1065,4 +1065,103 @@ function (l::MEGNetConv)(g::GNNGraph, x::AbstractMatrix, e::AbstractMatrix)
...
+    x = 1 / d .* x 
+
+    return l.σ(x .+ l.bias)
+end
+
+function Base.show(io::IO, l::GMMConv)
+    in, out, n_kernel, e_dim = l.ch[1], l.ch[2], l.n_kernel, l.e_dim
+    print(io, "GMMConv(", in, " => ", out)
+    print(io, ", n_kernel= ", n_kernel)
+    print(io, ", pseudo-cordinate dimension = ", e_dim)
```
CarloLucibello, line 1164:
```suggestion
    print(io, ", e_dim = ", e_dim)
```

src/layers/conv.jl
```diff
@@ -1065,4 +1065,103 @@ function (l::MEGNetConv)(g::GNNGraph, x::AbstractMatrix, e::AbstractMatrix)
...
+    e_dim::Int
+    dense_x::Dense
+end
+
+Flux.@functor GMMConv
+
+function GMMConv(ch::Pair{Int, Int}, 
+                n_kernel::Int,
+                e_dim::Int,
+                σ=identity;
```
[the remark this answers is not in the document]
melioristic: Is it a good idea to make e_dim default to 1? People might unknowingly try to use e of different dimensions and see errors.

melioristic:
> 

Test added as well

test/layers/conv.jl
```diff
@@ -262,4 +262,14 @@
...
         end
     end
+
+    @testset "GMMConv" begin
+        e_dim = 10
+        K = 5
+        l = GMMConv(in_channel => out_channel, K=K, e_dim=e_dim)
+        for g in test_graphs
+            g = GNNGraph(g, edata=rand(e_dim, g.num_edges))
+            test_layer(l, g, rtol=RTOL_HIGH, outsize = (out_channel, g.num_edges))
```
CarloLucibello, line 272:
```suggestion
            test_layer(l, g, rtol=RTOL_HIGH, outsize = (out_channel, g.num_nodes))
```

src/layers/conv.jl
```diff
@@ -1065,4 +1065,102 @@ function (l::MEGNetConv)(g::GNNGraph, x::AbstractMatrix, e::AbstractMatrix)
     return x̄, ē
 end
 
+@doc raw"""
+    GMMConv(in => out, K, e_dim, σ=identity; [init, bias])
+Graph mixture model convolution layer from the paper [Geometric deep learning on graphs and manifolds using mixture model CNNs](https://arxiv.org/abs/1611.08402)
```
CarloLucibello, line 1070:
```suggestion
    GMMConv(in => out, e_dim, σ=identity; [K, init, bias])

Graph mixture model convolution layer from the paper [Geometric deep learning on graphs and manifolds using mixture model CNNs](https://arxiv.org/abs/1611.08402).
```

CarloLucibello: Maybe the best compromise is to have `e_dim` has a mandatory positional argument and `K`  as keyword argument with default 1. Or we could have `e_dim` *mandatory* keyword arg. What do you think?  I don't want both of them to be positional since something like `GMMConv(16 => 16, 3, 5)`  is very unclear.

src/layers/conv.jl
```diff
@@ -1065,4 +1065,102 @@ function (l::MEGNetConv)(g::GNNGraph, x::AbstractMatrix, e::AbstractMatrix)
...
+w^a_{k}(e^a) = \exp(\frac{-1}{2}(e^a - \mu^a_k)^T \Sigma^a_k^{-1}(e^a - \mu^a_k))
+```
+$\Theta_k$, $\mu^a_k$, $\Sigma^a_k^{-1}$ are learnable parameters.
+
+The input to the layer is a node feature array 'X' of size `(num_features, num_nodes)` and
+edge pseudo-cordinate array 'U' of size `(num_features, num_edges)`
+# Arguments 
+- `in`: Number of input features.
+- `out`: Number of output features.
+- `K` : Number of kernels.
```
CarloLucibello, line 1086:
```suggestion
- `K` : Number of kernels (Default `1`).
```

src/layers/conv.jl
```diff
@@ -1065,4 +1065,102 @@ function (l::MEGNetConv)(g::GNNGraph, x::AbstractMatrix, e::AbstractMatrix)
...
+- `e_dim` : Dimensionality of pseudo coordinates.
```
CarloLucibello, line 1087:
```suggestion
- `e_dim` : Dimensionality of pseudo coordinates. Has to correspond to the edge features dimension in the forward pass.
```

src/layers/conv.jl
```diff
@@ -1065,4 +1065,102 @@ function (l::MEGNetConv)(g::GNNGraph, x::AbstractMatrix, e::AbstractMatrix)
...
+e = randn(e_dim, g.num_edges)
+
+# create layer
+l = GMMConv(in_feature=>out_feature, K, e_dim)
+
+# forward pass
+l(g, x, e)
+```
+"""
+
```
CarloLucibello, line 1110:
```suggestion
```
Otherwise the string won't be taken as a docstring

src/layers/conv.jl
```diff
@@ -1065,4 +1065,102 @@ function (l::MEGNetConv)(g::GNNGraph, x::AbstractMatrix, e::AbstractMatrix)
...
+    GMMConv(mu, sigma_inv, b, σ, ch, K, e_dim, dense_x)
+end
 
+function (l::GMMConv)(g::GNNGraph, x::AbstractMatrix, u::AbstractMatrix)
+
+    @assert (l.e_dim == size(u)[1] && g.num_edges == size(u)[2]) "Pseudo-cordinate dim $(size(u)) does not match (e_dim=$(e_dim),num_edge=$(g.num_edges))"
+
+    num_edges = g.num_edges
+    d = degree(g, dir=:in)
+    u = reshape(u, (l.e_dim, 1, num_edges))
```
CarloLucibello, line 1144: Let's respect the notation in the docstring formula. This should be called `e`, and what is currently called `e` should be called `w`.

src/layers/conv.jl
CarloLucibello: Actually, we should adopt the interface provided by `CGConv`  and have `GMMConv((in, ein) => out, act=identity; K=1, bias=true, init=glorot_uniform, residual=false)`.

test/layers/conv.jl
melioristic: Ahh My bad!

src/layers/conv.jl
melioristic: Working on it.

src/layers/conv.jl
```diff
@@ -1065,4 +1065,101 @@ function (l::MEGNetConv)(g::GNNGraph, x::AbstractMatrix, e::AbstractMatrix)
...
+end
+
+function (l::GMMConv)(g::GNNGraph, x::AbstractMatrix, u::AbstractMatrix)
+
+    @assert (l.e_dim == size(u)[1] && g.num_edges == size(u)[2]) "Pseudo-cordinate dim $(size(u)) does not match (e_dim=$(e_dim),num_edge=$(g.num_edges))"
+
+    num_edges = g.num_edges
+    d = degree(g, dir=:in)
+    u = reshape(u, (l.e_dim, 1, num_edges))
+    mu = reshape(l.mu, (l.e_dim, l.K, 1))
```
CarloLucibello, line 1144: mu is created as a `K x edim`, so the reshape should be `K x edim x 1`  to preserve natural ordering. Or mu should be created `edim x K`.

src/layers/conv.jl
```diff
@@ -1065,4 +1065,101 @@ function (l::MEGNetConv)(g::GNNGraph, x::AbstractMatrix, e::AbstractMatrix)
...
+    
+    e = -0.5*(u.-mu).^2
+    e = e .* ((reshape(l.sigma_inv, (l.e_dim, l.K, 1)).^2) )
```
CarloLucibello, line 1147:
```suggestion
    e = e .* reshape(l.sigma_inv, (l.e_dim, l.K, 1))
```
No need to square since `sigma_inv`  is an inverse covariance

src/layers/conv.jl
```diff
@@ -1065,4 +1065,101 @@ function (l::MEGNetConv)(g::GNNGraph, x::AbstractMatrix, e::AbstractMatrix)
...
+    e = exp.(sum(e, dims = 1 )) # (1, K, num_edge) 
+
+    xj = reshape(l.dense_x(x), (l.ch[2],l.K,:)) # (out, K, num_nodes)
```
CarloLucibello, line 1150:
```suggestion
    xj = reshape(l.dense_x(x), (l.ch[2], l.K, :)) # (out, K, num_nodes) 
```

melioristic: I did not get this part, I think it should still be K. Don't we need some transformation of x for each kernel K.

CarloLucibello: yes, you are right, ignore that

src/layers/conv.jl
melioristic: Added in the [commit](https://github.com/melioristic/GraphNeuralNetworks.jl/commit/510e4309b36e064ac42c441429e97b6990313d1d)

src/layers/conv.jl
melioristic: Thanks!

melioristic:
Added 
- (in, ein)=>out, with no default for ein, so the layer cannot be instantiated with GMMConv(in=>out), would prefer this way, but can discuss
- Added residual connection 
- Changed test accordingly
- Changed (cosmetic) the equation for weights, hope it looks understandable when rendered.

src/layers/conv.jl
```diff
@@ -1065,4 +1065,116 @@ function (l::MEGNetConv)(g::GNNGraph, x::AbstractMatrix, e::AbstractMatrix)
     return x̄, ē
 end
 
+@doc raw"""
+    GMMConv((in, ein) => out, σ=identity; K=1, bias=true, init=glorot_uniform, residual=false)
```
CarloLucibello, line 1069:
```suggestion
    GMMConv((in, ein) => out, σ=identity; K=1, bias=true, init=glorot_uniform, residual=false)

```

CarloLucibello: newline after signature

src/layers/conv.jl
```diff
@@ -1065,4 +1065,116 @@ function (l::MEGNetConv)(g::GNNGraph, x::AbstractMatrix, e::AbstractMatrix)
...
+# Arguments 
+
+- `in`: Number of input node features.
+- `ein`: Number of input edge features.
+- `out`: Number of output features.
+- `σ`: Activation function. Default `identity`.
+- `K`: Number of kernels. Default `1`.
+- `bias`: Add learnable bias. Default `true`.
+- `init`: Weights' initializer. Default `glorot_uniform`.
+- `residual`: Residual conncetion. Default `false`
```
CarloLucibello, line 1093:
```suggestion
- `residual`: Residual conncetion. Default `false`.
```

src/layers/conv.jl
```diff
@@ -1065,4 +1065,116 @@ function (l::MEGNetConv)(g::GNNGraph, x::AbstractMatrix, e::AbstractMatrix)
...
+function GMMConv(ch::Pair{NTuple{2,Int},Int}, 
+                σ=identity;
+                K::Int=1,
+                bias::Bool=true,
+                init=Flux.glorot_uniform,
+                residual=false)
+    
+    (nin, ein), out = ch
+    mu = init(ein, K)
+    sigma_inv = init(K, ein)
```
CarloLucibello, line 1135:
```suggestion
    sigma_inv = init(ein, K)
```

src/layers/conv.jl
```diff
@@ -1065,4 +1065,116 @@ function (l::MEGNetConv)(g::GNNGraph, x::AbstractMatrix, e::AbstractMatrix)
...
+    (nin, ein), out = l.ch #Notational Simplicity
+
+    @assert (ein == size(e)[1] && g.num_edges == size(e)[2]) "Pseudo-cordinate dim $(size(u)) does not match (ein=$(ein),num_edge=$(g.num_edges))"
+
+    num_edges = g.num_edges
+    d = degree(g, dir=:in)
+    w = reshape(e, (ein, 1, num_edges))
+    mu = reshape(l.mu, (ein, l.K, 1))
+    
+    w = -0.5*(w.-mu).^2
```
CarloLucibello, line 1152:
```suggestion
    w = -0.5 * (w .- mu).^2
```

src/layers/conv.jl
```diff
@@ -1065,4 +1065,116 @@ function (l::MEGNetConv)(g::GNNGraph, x::AbstractMatrix, e::AbstractMatrix)
...
+    w = w .* reshape(l.sigma_inv, (ein, l.K, 1))
+    w = exp.(sum(w, dims = 1 )) # (1, K, num_edge) 
+
+    xj = reshape(l.dense_x(x), (out,l.K,:)) # (out, K, num_nodes)
```
CarloLucibello, line 1156:
```suggestion
    xj = reshape(l.dense_x(x), (out, l.K, :)) # (out, K, num_nodes) 
```

src/layers/conv.jl
```diff
@@ -1065,4 +1065,116 @@ function (l::MEGNetConv)(g::GNNGraph, x::AbstractMatrix, e::AbstractMatrix)
...
+
+    xj = reshape(l.dense_x(x), (out,l.K,:)) # (out, K, num_nodes) 
+    m = propagate(e_mul_xj, g, +, xj=xj, e=w)
+    m = dropdims(mean(m, dims=2), dims=2) # (out, num_nodes)
+    m = 1 / d .* m 
+    
+    m = l.σ(m .+ l.bias)
+    
+    if l.residual
+        if size(x,1)==size(m,1)
```
CarloLucibello, line 1164:
```suggestion
        if size(x, 1) == size(m, 1)
```

src/layers/conv.jl
```diff
@@ -1065,4 +1065,116 @@ function (l::MEGNetConv)(g::GNNGraph, x::AbstractMatrix, e::AbstractMatrix)
...
+            m+=x
```
CarloLucibello, line 1165:
```suggestion
            m += x
```

CarloLucibello:
mostly stile comments, almost ready to go

src/layers/conv.jl
```diff
@@ -1065,4 +1065,117 @@ function (l::MEGNetConv)(g::GNNGraph, x::AbstractMatrix, e::AbstractMatrix)
...
+    sigma_inv = init(ein, K)
+    b = bias ? Flux.create_bias(ones(out), true) : false
+    dense_x = Dense(nin, out*K, bias=false)
+    GMMConv(mu, sigma_inv, b, σ, ch, K, dense_x, residual)
+end
+
+function (l::GMMConv)(g::GNNGraph, x::AbstractMatrix, e::AbstractMatrix)
+    (nin, ein), out = l.ch #Notational Simplicity
+
+    @assert (ein == size(e)[1] && g.num_edges == size(e)[2]) "Pseudo-cordinate dim $(size(u)) does not match (ein=$(ein),num_edge=$(g.num_edges))"
```
CarloLucibello, line 1145:
```suggestion
    @assert (ein == size(e)[1] && g.num_edges == size(e)[2]) "Pseudo-cordinate dim $(size(e)) does not match (ein=$(ein),num_edge=$(g.num_edges))"
```

melioristic:
> mostly stile comments, almost ready to go

Just being curious, do you use some kind of automatic formatter, any suggestions there.

CarloLucibello:
A first test failure could be solved by adding 
```
(l::GMMConv)(g::GNNGraph) = GNNGraph(g, ndata=l(g, node_features(g), edge_features(g)))
```

test/layers/conv.jl
```diff
@@ -262,4 +262,14 @@
                 outsize=((out_channel, g.num_nodes), (out_channel, g.num_edges))) 
         end
     end
+
+    @testset "GMMConv" begin
+        ein_channel = 10
+        K = 5
+        l = GMMConv((in_channel, ein_channel )=> out_channel, K=K)
+        for g in test_graphs
+            g = GNNGraph(g, edata=rand(ein_channel, g.num_edges))
```
CarloLucibello, line 271:
```suggestion
            g = GNNGraph(g, edata=rand(Float32, ein_channel, g.num_edges))
```
this should solve one of the test failures

melioristic:
I understand the error 1, this is that if my input is Float32 still my output is Float64, will see why this is happening. Could not understand the second test.

src/layers/conv.jl
```diff
@@ -1065,4 +1065,119 @@ function (l::MEGNetConv)(g::GNNGraph, x::AbstractMatrix, e::AbstractMatrix)
...
+                σ=identity;
+                K::Int=1,
+                bias::Bool=true,
+                init=Flux.glorot_uniform,
+                residual=false)
+    
+    (nin, ein), out = ch
+    mu = init(ein, K)
+    sigma_inv = init(ein, K)
+    b = bias ? Flux.create_bias(ones(out), true) : false
```
CarloLucibello, line 1137:
```suggestion
    b = bias ? Flux.create_bias(mu, true, out) : false
```
This should fix the Float64 issue

CarloLucibello:
> Could not understand the second test.

Me neither. Seems to be a Zygote error due to a not supported try/catch. But we already have `@assert` and `@warn` in other layers and they don't seem to cause problems. Maybe try to simplify the forward a bit and test locally with something like
```julia
l = GMMConv((2,2)=>2)
g = rand_graph(5, 10, ndata=rand(Float32,2,5), edata=rand(Float32,2,10))
gradient(() -> sum(l(g, g.ndata.x, g.edata.e)), Flux.params(l))  
```

src/layers/conv.jl
melioristic: The problem is here. Both w and mu are Float32 before this, but it changes to Float64 at this point. How exactly `@.` functions? `w = @. -0.5 * (w - mu)^2`

melioristic:
> > Could not understand the second test.
> 
> Me neither. Seems to be a Zygote error due to a not supported try/catch. But we already have `@assert` and `@warn` in other layers and they don't seem to cause problems. Maybe try to simplify the forward a bit and test locally with something like
> 
> ```julia
> l = GMMConv((2,2)=>2)
> g = rand_graph(5, 10, ndata=rand(Float32,2,5), edata=rand(Float32,2,10))
> gradient(() -> sum(l(g, g.ndata.x, g.edata.e)), Flux.params(l))  
> ```

It is probably related to the string interpolation error in gradient. But changing this
`"Pseudo-cordinate dim $(size(e)) does not match (ein=$(ein),num_edge=$(g.num_edges))"`
to 
`"Pseudo-cordinate dim does not match (ein, num_edge))"`
runs everything smoothly

src/layers/conv.jl
melioristic: I don't know if this is the most elegant solution, but I had to change the code at two different points. 
- `w = @. -0.5 * (w - mu)^2` goes to `w = @. - ((w - mu) ^ 2)/ 2`
- ` m = 1 / d .* m ` goes to `m = m./reshape(d, (1, g.num_nodes))`
The problem is if in degrees say d=[1,2,1] is operated to 1/d then it converts to Float 64, depending on the system. Same with 2 and 0.5 I guess.

src/layers/conv.jl
```diff
@@ -1065,4 +1065,121 @@ function (l::MEGNetConv)(g::GNNGraph, x::AbstractMatrix, e::AbstractMatrix)
...
+end
+
+function (l::GMMConv)(g::GNNGraph, x::AbstractMatrix, e::AbstractMatrix)
+    (nin, ein), out = l.ch #Notational Simplicity
+
+    @assert (ein == size(e)[1] && g.num_edges == size(e)[2]) "Pseudo-cordinate dimension is not equal to (ein,num_edge)"
+
+    num_edges = g.num_edges
+    d = degree(g, dir=:in)
+    print(eltype(d))
```
CarloLucibello, line 1149:
```suggestion
```

CarloLucibello, line 1149:
```suggestion
```

melioristic:
NANs possibly because dividing by indegree, instead changed the aggregator method to mean. This should potentially get. rid of the NAN values

src/layers/conv.jl
```diff
@@ -1065,4 +1065,118 @@ function (l::MEGNetConv)(g::GNNGraph, x::AbstractMatrix, e::AbstractMatrix)
...
+    (nin, ein), out = l.ch #Notational Simplicity
+
+    @assert (ein == size(e)[1] && g.num_edges == size(e)[2]) "Pseudo-cordinate dimension is not equal to (ein,num_edge)"
+
+    num_edges = g.num_edges
+    w = reshape(e, (ein, 1, num_edges))
+    mu = reshape(l.mu, (ein, l.K, 1))
+    
+    w = @. ((w - mu)^2) / 2
+    w = w .* reshape(l.sigma_inv, (ein, l.K, 1))
```
CarloLucibello, line 1152:
```suggestion
    w = w .* reshape(l.sigma_inv.^2, (ein, l.K, 1))
```
So `sigma_inv` models an inverse standard deviation and we guarantee that the inverse variance is non-negative as it should be

src/layers/conv.jl
```diff
@@ -1065,4 +1065,118 @@ function (l::MEGNetConv)(g::GNNGraph, x::AbstractMatrix, e::AbstractMatrix)
...
+                            
+(l::GMMConv)(g::GNNGraph) = GNNGraph(g, ndata=l(g, node_features(g), edge_features(g)))
+
+function Base.show(io::IO, l::GMMConv)
+    (nin, ein), out = l.ch
+    print(io, "GMMConv((", nin, ",", ein, ")=>", out)
+    print(io, ", K=", l.K)
+    print(io, ", σ=", l.σ)
+    print(io, ")")
+
```
CarloLucibello, line 1181:
```suggestion
```

src/layers/conv.jl
```diff
@@ -1065,4 +1065,118 @@ function (l::MEGNetConv)(g::GNNGraph, x::AbstractMatrix, e::AbstractMatrix)
...
+    return m
+end
+                            
+(l::GMMConv)(g::GNNGraph) = GNNGraph(g, ndata=l(g, node_features(g), edge_features(g)))
+
+function Base.show(io::IO, l::GMMConv)
+    (nin, ein), out = l.ch
+    print(io, "GMMConv((", nin, ",", ein, ")=>", out)
+    print(io, ", K=", l.K)
+    print(io, ", σ=", l.σ)
```
CarloLucibello, line 1179: this should be printed before K and only if different from identity

melioristic: So sigma should be printed before k and only if different from identity? Changing that.

CarloLucibello:
Great, impressive works, thanks!

melioristic:
> Great, impressive works, thanks!

Thanks a lot for all the help! Will try some other conv layer this weekend. :) And then will see about temporal GNN.

PR #154, opened 2022-04-09
SGConv

rbSparky:
Adding SGConv from the paper [Semi-Supervised Classification with Graph Convolutional Networks](https://arxiv.org/pdf/1902.07153.pdf).


rbSparky:
What I understand from the paper is that we create a layer, which has the same feature propagation with the normalised adjency matrix step as ```GCNConv```, but for the ```k``` hops, we propagate k times (which can be done with a loop?), and then in the end, we return the return value without applying the activation function.

Can you confirm if I have the right idea? I will try to test it and then make the next commit.

CarloLucibello:
You put the wrong reference in the opening post, the correct one is https://arxiv.org/pdf/1902.07153.pdf

CarloLucibello:
> What I understand from the paper is that we create a layer, which has the same feature propagation with the normalised adjency matrix step as GCNConv, but for the k hops, we propagate k times (which can be done with a loop?), and then in the end, we return the return value without applying the activation function.

Yes that's correct. Instead of copying over the implementation of GCNConv forward pass, we could actually store a GCNConv with no activation inside of the SGConv and use that K times in the forward pass.

rbSparky:
> Yes that's correct. Instead of copying over the implementation of GCNConv forward pass, we could actually store a GCNConv with no activation inside of the SGConv and use that K times in the forward pass.

Makes sense, I'll do that.

> You put the wrong reference in the opening post, the correct one is https://arxiv.org/pdf/1902.07153.pdf

Oh yes, my apologies, I've fixed that now.

rbSparky:
Just a small question, after removing the non linearity we have \hat{Y} = softmax(S^{k} X \Theta^{(1)}\Theta^{(2)} \dots \Theta^{(k)}) but in we collapse the linear transformation into only one of such \hat{Y} = softmax(S^{k}X \Theta), if we just use GCNConv ```k``` times in the forward pass, wouldn't we be implementing the first equation?

CarloLucibello:
ouch sorry, I was mistaken, applying the GCNConv multiple times is not equivalent to SGConv since we don't want to multiple by the weights multiple times but only once. Sorry for derailing you!

rbSparky:
I think we should now move on to testing this? Should I write some tests similar to the ones for other layers, since I'm not very sure how the specifics of this would be tested though(like the final result for example, or if we even need to test that). What do you suggest?

CarloLucibello:
The implementation looks good. For testing you can follow the example of the other layers. We test output size, equality between cpu an gpu outputs, differentiability.

rbSparky:
I've added the tests, is there anything else that needs to be done before merging?

src/layers/conv.jl
```diff
@@ -1181,3 +1181,120 @@ function Base.show(io::IO, l::GMMConv)
     l.residual==true || print(io, ", residual=", l.residual)
     print(io, ")")
 end
+
+@doc raw"""
+    SGConv(int => out, k=1; [bias, init, add_self_loops, use_edge_weight])
```
CarloLucibello, line 1186:
```suggestion
    SGConv(int => out, k=1; [bias, init, add_self_loops, use_edge_weight])
    
```

src/layers/conv.jl
```diff
@@ -1181,3 +1181,120 @@ function Base.show(io::IO, l::GMMConv)
...
+y = l(g, x, w)
+
+# Edge weights can also be embedded in the graph.
+g = GNNGraph(s, t, w)
+l = SGConv(3 => 5, add_self_loops = true, use_edge_weight=true) 
+y = l(g, x) # same as l(g, x, w) 
+```
+"""
+
+struct SGConv{A<:AbstractMatrix, B} <: GNNLayer
```
CarloLucibello, line 1231:
```suggestion
"""
struct SGConv{A<:AbstractMatrix, B} <: GNNLayer
```

rbSparky: Done

CarloLucibello (approved):

CarloLucibello:
thanks for this!

PR #158, opened 2022-04-15
Added EdgeWeightNorm layer

rbSparky:
Verified outputs with the pytorch implementation

### DIFF
--- a/src/layers/conv.jl
+++ b/src/layers/conv.jl
@@ -1181,3 +1181,69 @@ function Base.show(io::IO, l::GMMConv)
     l.residual==true || print(io, ", residual=", l.residual)
     print(io, ")")
 end
+
+@doc raw"""
+    EdgeWeightNorm(norm_both = true, eps = 0)
+
+Normalizes positive scalar edge weights on a graph following the form in GCN.
+
+norm_both = `true` yields the following normalization term:
+```math
+c_{ji} = (\sqrt{\sum_{k\in\mathcal{N}(j)}e_{jk}}\sqrt{\sum_{k\in\mathcal{N}(i)}e_{ki}})
+```
+norm_both = `false` yields the following normalization term:
+```math
+c_{ji} = (\sum_{k\in\mathcal{N}(i)}e_{ki})
+```
+where ``e_ji`` is the scalar weight on the edge from node j to node i.
+
+Return value is the normalized weight ``e_{ji} / c_{ji}`` for all edges in vector form.
+
+# Arguments
+
+- `norm_both`: The normalizer as specified above. Default is `true`.
+- `eps`: Offset value in the denominator. Default is `0`.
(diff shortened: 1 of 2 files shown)


src/layers/conv.jl
```diff
@@ -1181,3 +1181,69 @@ function Base.show(io::IO, l::GMMConv)
...
+
+@functor EdgeWeightNorm
+
+function EdgeWeightNorm(norm_both::Bool = true, 
+                        eps::Float64 = 0)                  
+    EdgeWeightNorm(norm_both, eps)
+end
+
+function (l::EdgeWeightNorm)(g::GNNGraph, edge_weight::AbstractVector)
+    norm_val = Vector{Float64}()
```
CarloLucibello, line 1234: the output type should be of the same type of  `edge_weight`. We want to support cuda arrays and Float32 operations.

CarloLucibello, line 1234: the output type should be of the same type of  `edge_weight`. We want to support cuda arrays and Float32 operations.

src/layers/conv.jl
```diff
@@ -1181,3 +1181,69 @@ function Base.show(io::IO, l::GMMConv)
...
+function (l::EdgeWeightNorm)(g::GNNGraph, edge_weight::AbstractVector)
+    norm_val = Vector{Float64}()    
+    edge_in, edge_out = edge_index(g)
+    
+    dg_in = degree(g; dir = :in, edge_weight)
+    dg_out = degree(g; dir = :out, edge_weight)
+    
+    for iter in 1:length(edge_weight)
+        if l.norm_both
+            push!(norm_val, edge_weight[iter] / (sqrt(dg_out[in[iter]] * dg_in[out[iter]]) + l.eps))
```
CarloLucibello, line 1242: these mutating operations are not AD friendly. I didn't think about it carefully but you should probably use [apply_edges](https://carlolucibello.github.io/GraphNeuralNetworks.jl/dev/api/messagepassing/#GraphNeuralNetworks.apply_edges) here

CarloLucibello, line 1242: these mutating operations are not AD friendly. I didn't think about it carefully but you should probably use [apply_edges](https://carlolucibello.github.io/GraphNeuralNetworks.jl/dev/api/messagepassing/#GraphNeuralNetworks.apply_edges) here

rbSparky: I tried doing this using aggregate_neighbours, but ran into the same issue of it not being AD friendly, main issue being that I'm currently computing `∑e_jk` and `∑e_ki` individually for each `e_ji`, which I guess is to be done with some sort of matrix multiplication. 

I'm not sure yet how its to be done using apply_edges, but I'll look more into it(will try to understand how its done in pyTorch/DGL) and let you know any updates.

PR #159, opened 2022-04-15
GlobalConcatPool

casper2002casper:
Concatenates global features to the node features, is used to integrate global properties into the output of the nodes.

### DIFF
--- a/src/layers/pool.jl
+++ b/src/layers/pool.jl
@@ -44,6 +44,25 @@ end
 
 (l::GlobalPool)(g::GNNGraph) = GNNGraph(g, gdata=l(g, node_features(g)))
 
+@doc raw"""
+    GlobalConcatPool(aggr)
+
+```math
+\mathbf{x}_i' = [\mathbf{x}_i; \square_{i \in V} \mathbf{x}_i]
+
+"""
+struct GlobalConcatPool{F} <: GNNLayer
+    aggr::F
+end
+  
+function (l::GlobalConcatPool)(g::GNNGraph, x::AbstractArray)
+    g_feat = reduce_nodes(l.aggr, g, x)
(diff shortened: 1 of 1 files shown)


src/layers/pool.jl
```diff
@@ -44,6 +44,25 @@ end
 
 (l::GlobalPool)(g::GNNGraph) = GNNGraph(g, gdata=l(g, node_features(g)))
 
+@doc raw"""
+    GlobalConcatPool(aggr)
+
+```math
+\mathbf{x}_i' = [\mathbf{x}_i; \square_{i \in V} \mathbf{x}_i]
+
```
CarloLucibello, line 52:
```suggestion
```math
\mathbf{x}_i' = [\mathbf{x}_i; \square_{j \in V} \mathbf{x}_j]
```
```

File: src/layers/pool.jl [hunk 2]
```diff
@@ -44,6 +44,25 @@ end
 
 (l::GlobalPool)(g::GNNGraph) = GNNGraph(g, gdata=l(g, node_features(g)))
 
+@doc raw"""
+    GlobalConcatPool(aggr)
+
```
CarloLucibello, line 49: needs a docstring

src/layers/pool.jl
CarloLucibello: also the math block is not closed by ```

CarloLucibello:
Nice addition, thanks. Needs tests and a few words explaining the behavior in the docstring.

casper2002casper:
I changed it a bit, so any pooling layer can be used. Do you agree with this change?
If so I'll some documentation

casper2002casper:
The test uses a GlobalPool layer, is this OK as it happens after the GlobalPool test, or should the test be fully isolated?

PR #172, opened 2022-05-23
add Pluto tutorials

### LINKED ISSUE #168: Add tutorials written in Pluto
We can port these tutorials to GNN.jl:
- https://pytorch-geometric.readthedocs.io/en/latest/notes/colabs.html
- https://antoniolonga.github.io/Pytorch_geometric_tutorials/

To set up Pluto + Documenter, see this repo:
https://github.com/JuliaManifolds/Manopt.jl

CarloLucibello:
A step toward fixing #168.

Also adding a few utilities
-  `Graphs.Graph(::GNNGraph)` conversion
- `has_isolated_nodes(g)`
- `to_unidirected(g)` which removes the reves edges and can be useful for plotting an undirected graph which is bidirected as an undirected one.


CarloLucibello:
There are some problems that I don't understand in the github action compiling the pluto notebooks. I'll leave them out for the time being and solve in a followup PR

rikhuijzer:
> There are some problems that I don't understand in the github action compiling the pluto notebooks. I'll leave them out for the time being and solve in a followup PR

Cool to see that you're going to use PlutoStaticHTML. Let's get it working

You have set
```julia
use_distributed=false
```
From the PlutoStaticHTML docs:
> Beware that `use_distributed=false` will not work with Pluto's built-in package manager.

So, you can fix it in two ways. The first option is to explicitly install the packages manually. So, something like:
```julia
begin
    using Pkg
    Pkg.activate(; temp=true)
    packages = [
        PackageSpec(; path=dirname(dirname(@__DIR__))),
        PackageSpec(; name="Flux", version="0.13")
        [...]
    ]
    Pkg.add(packages)
end
```
```julia
begin
    using GraphNeuralNetworks
    using Flux
    ...
end
```
I would personally do this because it is very explicit and therefore easy to maintain.

An alternative would be to set `use_distributed=true` and use Pluto's built-in package manager. However, then you need to figure out how to install the latest GraphNeuralNetworks in your notebook. This is particularly difficult when you are on a PR. Or do you know a way to install a relative path via Pluto's package manager? Maybe that's possible. I haven't checked in detail

PR #174, opened 2022-05-25
fixing CI for pluto tutorials

CarloLucibello:


docs/make.jl
```diff
@@ -87,11 +88,11 @@ makedocs(;
...
             #  "Tutorials" => tutorial_menu,
-            #  "Tutorials" => 
-            #     [
-            #         "Intro to Graph Neural Networks" => "gnn_intro_pluto.md",
-            #         "Graph Classification" => "graph_classification_pluto.md",
-            #     ],
+             "Tutorials" => 
+                [
+                    "Intro to Graph Neural Networks" => "gnn_intro_pluto.md",
+                    "Graph Classification" => "graph_classification_pluto.md",
```
rikhuijzer:
```suggestion
                    "Intro to Graph Neural Networks" => "tutorials/gnn_intro_pluto.md",
                    "Graph Classification" => "tutorials/graph_classification_pluto.md",
```

rikhuijzer:
@CarloLucibello In `graph_classification_pluto`, don't forget to set
```julia
dataset = let
	ENV["DATADEPS_ALWAYS_ACCEPT"] = "true"
	TUDataset("MUTAG")
end
```
without that, the build fails locally here so probably also in CI

docs/src/tutorials/gnn_intro_pluto.jl
```diff
@@ -4,6 +4,22 @@
...
+        PackageSpec(; path=joinpath(@__DIR__,"..","..","..")),
+        PackageSpec(; name="Flux", version="0.13"),
+		PackageSpec(; name="MLDatasets", version="0.7"),
+		PackageSpec(; name="GraphMakie"),
+		PackageSpec(; name="Graphs"),
+		PackageSpec(; name="CairoMakie"),
+		PackageSpec(; name="PlutoUI"),
+    ]
+    Pkg.add(packages)
+end
```
CarloLucibello, line 21: @rikhuijzer a problem with this is that I get a very long terminal output in the notebook that I don't know how to suppress

CarloLucibello: maybe I just add all those packages to `docs/Project.toml` and activate that?

rikhuijzer: In Pluto you mean right? In the few latest versions you can disable console output per cell. Select it at the top right of the cell. It remembers your choice also between runs

CarloLucibello:
hurrah!

docs/src/tutorials/gnn_intro_pluto.jl
rikhuijzer: And if you want to hide the code for a cell, use `# hideall` like in Documenter:

```julia
# hideall
x = 1
```
will show only the output block containing 1. To hide code and output, use the semicolon.

rikhuijzer:
> hurrah!

:partying_face: Looks like it works.

Can I add `GraphNeuralNetworks.jl` to the list of packages using `PlutoStaticHTML.jl` for documentation (https://huijzer.xyz/PlutoStaticHTML.jl/dev/#Documenter.jl)? You're also welcome to open a PR if you want

docs/src/tutorials/gnn_intro_pluto.jl
CarloLucibello: what happens with the visibility toggle in Pluto frontend? Isn't the visibility state stored in the cell metadata?

rikhuijzer: Yep nowadays it is. I have to implement reading that still. Added it to my task list a few minutes ago (https://github.com/rikhuijzer/PlutoStaticHTML.jl/issues/119)

rikhuijzer: > Isn't the visibility state stored in the cell metadata?

I've tested this now but apparently not no in 0.19.5

PR #176, opened 2022-05-27
Documentation fixes for GATv2Conv

### LINKED ISSUE #175: Inaccurate `GATv2Conv` Documentation
The documentation on `GATv2Conv` says:

<blockquote>
the attention coefficients $\alpha_{ij}$ are given by

$$\alpha_{ij} = \frac{1}{z_i} \exp(\mathbf{a}^T LeakyReLU([W_2 \mathbf{x}_i; W_1 \mathbf{x}_j]))$$
</blockquote>

This is not exactly the same as the Equation (7) from “How Attentive are Graph Attention Networks?”:

$$e(\mathbf{h}_i, \mathbf{h}_j) = \mathbf{a}^T LeakyReLU(\mathbf{W} \cdot [\mathbf{h}_i || \mathbf{h}_j])$$

I also checked the relevant piece of code (`conv.jl`):

```
    function message(Wix, Wjx, e)
        Wx = Wix + Wjx
        # [...]
        eij = sum(l.a .* leakyrelu.(Wx, l.negative_slope), dims=1)   # 1 × heads × nedges
        α = exp.(eij)
        return (α = α, β = α .* Wjx)
    end
```

Thanks to my project supervisor, I now realize that this is all correct and equivalent to the paper definition.
So it appears that the issue lies with the documentation.
Instead of a vector concatenation, it should state addition, or stick with the original order of operations.

As a small side note, the markup on `GATConv` is a bit broken, and it says “D*a*fault” instead of “Default”.

Animiral:
As well as the change to addition, added a code comment about equivalence to the GATv2 paper.

Resolve #175.

### DIFF
--- a/src/layers/conv.jl
+++ b/src/layers/conv.jl
@@ -275,7 +275,7 @@ In case `ein > 0` is given, edge features of dimension `ein` will be expected in
 and the attention coefficients will be calculated as  
 ```math
 \alpha_{ij} = \frac{1}{z_i} \exp(LeakyReLU(\mathbf{a}^T [W_e \mathbf{e}_{j\to i}; W \mathbf{x}_i; W \mathbf{x}_j]))
-````
+```
 
 # Arguments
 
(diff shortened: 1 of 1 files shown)


CarloLucibello:
thanks!

Issue #184, opened 2022-06-08
Spelling Error of edge?


umbriquse:
https://github.com/CarloLucibello/GraphNeuralNetworks.jl/blob/78792f0978f047dd36c12ac65bd9555de21e53c8/src/layers/conv.jl#L404

CarloLucibello:
right, thanks for reporting!

CarloLucibello:
Closed in #186

PR #186, opened 2022-06-08
Fix Spelling

umbriquse:


umbriquse:
Proposed Fix to #184

CarloLucibello:
ahah, we filed a PR in sync. I'll close mine. Btw, same mistake is on line 283

Issue #200, opened 2022-07-21
Flux destructure/restructure bug


mplemay:
Hi all, when when destructing and then restructuring a GNN chain I get the following error.

```julia
julia> using GraphNeuralNetworks, Flux, Pkg

julia> Pkg.status()
Project GNNBug v0.1.0
Status `~/Github/GNNBug.jl/Project.toml`
  [587475ba] Flux v0.13.4
  [cffab07f] GraphNeuralNetworks v0.4.4

julia> chain = GNNChain(GraphConv(2=>2))
GNNChain(GraphConv(2 => 2, aggr=+))

julia> params, restructure = Flux.destructure(chain)
(Float32[0.70314914, 0.8151582, 0.8936175, 0.8738015, 0.17640713, -0.52357626, -0.92551565, 0.015487637, 0.0, 0.0], Restructure(GNNChain, ..., 10))

julia> restructure(params)
ERROR: type Tuple has no field layers
Stacktrace:
 [1] getproperty
   @ ./Base.jl:38 [inlined]
 [2] functor(#unused#::Type{GNNChain{Tuple{GraphConv{Matrix{Float32}, Vector{Float32}}}}}, c::Tuple{NamedTuple{(:weight1, :weight2, :bias, :σ, :aggr), Tuple{Int64, Int64, Int64, Tuple{}, Tuple{}}}})
   @ GraphNeuralNetworks ~/.julia/packages/GraphNeuralNetworks/4e2gU/src/layers/basic.jl:117
 [3] _trainable_biwalk(f::Function, x::GNNChain{Tuple{GraphConv{Matrix{Float32}, Vector{Float32}}}}, aux::Tuple{NamedTuple{(:weight1, :weight2, :bias, :σ, :aggr), Tuple{Int64, Int64, Int64, Tuple{}, Tuple{}}}})
   @ Optimisers ~/.julia/packages/Optimisers/AqvxP/src/destructure.jl:95
 [4] #fmap#30
   @ ~/.julia/packages/Functors/qBIlC/src/functor.jl:78 [inlined]
[...]

PR #201, opened 2022-07-21
fix GNNChain restructure bug

mplemay:
Hi this is my attempted fix for #200. Please let me know if there is anything I can do to improve this PR.

### DIFF
--- a/test/layers/basic.jl
+++ b/test/layers/basic.jl
@@ -64,5 +64,11 @@
         wg = WithGraph(model, g, traingraph=true)
         @test length(Flux.params(wg)) == length(Flux.params(model)) + length(Flux.params(g))
     end
+
+    @testset "Flux restructure" begin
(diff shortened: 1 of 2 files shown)


test/layers/basic.jl
```diff
@@ -64,5 +64,11 @@
         wg = WithGraph(model, g, traingraph=true)
         @test length(Flux.params(wg)) == length(Flux.params(model)) + length(Flux.params(g))
     end
+
+    @testset "Flux restructure" begin
+        chain = GNNChain(GraphConv(2=>2))
+        params, restructure = Flux.destructure(chain)
+        restructure(params)
```
CarloLucibello, line 71:
```suggestion
        @test restructure(params) isa GNNChain
```

mplemay: Good Idea. Thank you for the fix.

PR #202, opened 2022-07-22
Update GNNChain

CarloLucibello:
Keeping in sync GNNChain with last months' changes with Flux.Chain implementation (https://github.com/FluxML/Flux.jl/pull/1809)

TODO:
- [x] improve GNNChain docstring
- [ ] benchmarks

PR #203, opened 2022-07-22
Replace depreciated Flux.ADAM with Flux.Adam

mplemay:
This PR replaces the depreciated `Flux.ADAM` optimizer with `Flux.Adam` across the entire project. This PR only changes, docs, performance benchmarks, tests, and examples so it should it should have not break end-user code.


mplemay:
@CarloLucibello I just noticed you created another PR that addresses this. I will close this.

CarloLucibello:
I actually I would prefer if we merge this PR instead of having an unrelated change in another PR

CarloLucibello:
We need to update Flux compat to 0.13.4

mplemay:
> We need to update Flux compat to 0.13.4

@CarloLucibello I am relatively new to Julia. In this case would it make more sense to set the compatibility to `^0.13.4` or `~0.13.4`?

CarloLucibello:
See how it is done in #202

PR #205, opened 2022-09-17
Fixed a few typos in notebooks, docs and docstrings

pitmonticone:


CarloLucibello:
mhmhm, CI is not passing, looks like some recent change in ChainRules caused some breackage. 

Thanks for this PR!

PR #206, opened 2022-09-29
Switch to DemoCards.jl

Dsantra92:
This pr sets DemoCards for the tutorials page, removing direct Pluto and StaicHTML dependency.

Also added cache files for reducing time on CI. With a few more notebooks, running CI on notebooks would be very costly.

Merge after [DemoCards.jl pluto pr](https://github.com/JuliaDocs/DemoCards.jl/pull/121)

Screenshots:
screenshot1
screenshot

PR #207, opened 2022-09-29
Fix the current failing tests

Dsantra92:
- [x] Edge a Graph
- [ ] NNConv gradients
Docs are failing due to PlutoStaitcHTML, will be fixed with #206 .


CarloLucibello:
The problem with Graph.jl has been fixed in https://github.com/JuliaGraphs/Graphs.jl/pull/174. 
I'm working on the NNConv right now.

Dsantra92:
Okay, closing it then.

PR #230, opened 2022-11-22
Cache Pluto Notebooks

Dsantra92:
Given the implementation in DemoCards.jl we should be able to support cache notebook renders based on the input notebook's sha. Unfortunately, we haven't been able to use that feature because the Julia version that we are using for generating docs on CI caused a mismatch. I am updating the CI's Julia version add adding updated cache renders. Let's see how it goes. 

If everything goes right, the docs build should use the cache files instead of evaluating the notebook files.


Dsantra92:
Looks like the cache is working. See [docs build](https://github.com/CarloLucibello/GraphNeuralNetworks.jl/actions/runs/3521732344/jobs/5903905252).

CarloLucibello:
It's working! Doc build down to 1m30s, used to be 30 min. Thanks!

PR #232, opened 2022-11-28
implement DataStore

CarloLucibello:
Provide more flexibility to GNNGraph by storing features into a custom class, the `DataStore`, with an interface inheriting from both dictionaries and named tuples. We can now mutate the graph storing new features arrays with
```
g.ndata.z = rand(3, 10)
```


CarloLucibello:
tests are passing locally but for some reason they are not passing on the CI

PR #241, opened 2022-12-07
Fix gnnheterographs failed test

aurorarossi:
With this PR I fixed the print order of `Dict`s and the gnnheterographs tests  should now not fail.

### DIFF
--- a/src/GNNGraphs/gnnheterograph.jl
+++ b/src/GNNGraphs/gnnheterograph.jl
@@ -143,13 +143,13 @@ function Base.show(io::IO, ::MIME"text/plain", g::GNNHeteroGraph)
         g.num_graphs > 1 && print(io, "\n num_graphs: $(g.num_graphs)")
         if !isempty(g.ndata)
             print(io, "\n ndata:")
-            for k in keys(g.ndata)
+            for k in sort!(collect(keys(g.ndata)))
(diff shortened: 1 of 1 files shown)


aurorarossi:
(I will open another PR for the `laplacian_lambda_max` error)

src/GNNGraphs/gnnheterograph.jl
```diff
@@ -130,26 +130,58 @@ function GNNHeteroGraph(data::EDict;
         ntypes, etypes)
 end
 
+function show_sorted_Dict(io::IO, d::Dict,compact::Bool)
```
CarloLucibello, line 133:
```suggestion
function show_sorted_Dict(io::IO, d::Dict, compact::Bool)
```
and similar syntax changes in all the added code

src/GNNGraphs/gnnheterograph.jl
```diff
@@ -130,26 +130,58 @@ function GNNHeteroGraph(data::EDict;
...
+    if compact
+        print(io,"Dict")
+    end
+    print(io,"(")
+    if !isempty(d)
+        if length(keys(d))==1
```
CarloLucibello, line 139:
```suggestion
        if length(keys(d)) == 1
```

PR #242, opened 2022-12-13
Fix test laplacian lambda max

aurorarossi:
With this PR I added the seed for `rand_graph`.

### DIFF
--- a/test/GNNGraphs/query.jl
+++ b/test/GNNGraphs/query.jl
@@ -135,12 +135,14 @@
     @testset "laplacian_lambda_max" begin
         s = [1, 2, 3, 4, 5, 1, 2, 3, 4, 5]
         t = [2, 3, 4, 5, 1, 5, 1, 2, 3, 4]
-        g = GNNGraph(s,t)
+        g = GNNGraph(s, t)
         @test laplacian_lambda_max(g) ≈ Float32(1.809017)
-        data = [g for i in 1:5]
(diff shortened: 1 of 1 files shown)


test/GNNGraphs/query.jl
```diff
@@ -137,12 +137,10 @@
...
-        data1 = [g for i in 1:5]
-        gall1 = Flux.batch(data1)
-        @test laplacian_lambda_max(gall1) ≈ [Float32(1.809017) for i in 1:5]
-        data2 = [rand_graph(10,20) for i in 1:3]
-        gall2 = Flux.batch(data2)
-        @test length(laplacian_lambda_max(gall2)) == 3
+        data = [g for i in 1:5]
+        gall = Flux.batch(data)
+        @test laplacian_lambda_max(gall) ≈ [Float32(1.809017) for i in 1:5]
+        @test length(laplacian_lambda_max(gall)) == 5
```
CarloLucibello, line 143:
```suggestion
```
the length is already implicitly tested in the line above

aurorarossi:
Yes and also in the function `_eigmax` there is the following line:
```jl
 KrylovKit.eigsolve(Symmetric(A), x0, 1, :LR)[1][1]
```
hence A is forced to be symmetric. 
When I run the tests on my computer they run smoothly (I have Julia 1.8).

CarloLucibello:
So all this PR does is remove a test...

PR #244, opened 2022-12-23
fixing overflow in GAT layers

pevnak:
I have found that attention in GAT can overflow, as attention layers likes to due to the presence of exp which is not checked for `maximum`. I propose a quick and likely very dirty hack. What I do not like on the solution is that it done in function `message` and works only because there is a subsequent division in other function. I run full test suit and seems to work.

### DIFF
--- a/src/layers/conv.jl
+++ b/src/layers/conv.jl
@@ -367,7 +367,9 @@ function message(l::GATConv, Wxi, Wxj, e)
         Wxx = vcat(Wxi, Wxj, We)
     end
     aWW = sum(l.a .* Wxx, dims=1)   # 1 × nheads × nedges
-    α = exp.(leakyrelu.(aWW, l.negative_slope))       
+    logα = leakyrelu.(aWW, l.negative_slope)
+    maxα = maximum(logα)
+    α = exp.(logα .- maxα)       
     return (α = α, β = α .* Wxj)
 end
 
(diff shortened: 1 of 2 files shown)


CarloLucibello:
A problem of the maximum trick in this setting is that the maximum is computed over all edges in the graph, 
while the softmax is over the neighborhood. 
I fear it could cause underflow problems instead of overflow ones.

A robust solution involves [softmax_edges_neighborhood](https://github.com/CarloLucibello/GraphNeuralNetworks.jl/blob/77f0e0f6807636163eea239b03e8840e9916dc0f/src/utils.jl#L70), which is also what PyG uses in GAT layers.
That function should also be modified to add an eps to the denominator.

pevnak:
I see, 

but we agree that the implementation of GATConv has overflow issue and should be fixed, right?

pevnak:
A new attempt to fix it, though I had to write `propagate` to the `(l::GATConv)`

CarloLucibello:
Yes it should be fixed, but likely we will lose a factor close to two in performance

pevnak:
I know, but do not know what to do with this. Seems like it is done either fast or correctly. Do you have any idea, how to deal with it?

CarloLucibello:
let's be safe, it's also what other frameworks do. Can you apply the same pattern to `GATv2Conv`?

PR #249, opened 2022-12-28
Add a transformer-like convolutional layer

graidl:
I suggest to add a transformer-like convolutional layer. The proposed implementation follows [`TransformerConv` from Pytorch Geometric](https://pytorch-geometric.readthedocs.io/en/latest/modules/nn.html#torch_geometric.nn.conv.TransformerConv) but also contains a few additional options. More specifically, the transformer-like multi head attention convolutional operator from the [Masked Label Prediction: Unified Message Passing Model for Semi-Supervised Classification](https://arxiv.org/abs/2009.03509) paper with optional edge features is realized. Additionally, there are options to configure it as the transformer-like convolutional operator from the [Attention, Learn to Solve Routing Problems!](https://arxiv.org/abs/1803.08475) paper, including a successive feed-forward network as well as skip layers and batch normalization.

Carlo: edited with the correct link to the Attention, Learn to solve... paper

### DIFF
--- a/src/layers/conv.jl
+++ b/src/layers/conv.jl
@@ -1545,8 +1545,7 @@ function TransformerConv(ch::Pair{NTuple{2, Int}, Int};
     (in, ein), out = ch
 
     if add_self_loops
-        @assert iszero(ein) "Using edge features and setting add_self_loops=true at the \
-            same time is not yet supported."
+        @assert iszero(ein) "Using edge features and setting add_self_loops=true at the same time is not yet supported."
     end
 
     W1 = root_weight ? Dense(in, out * (concat ? heads : 1); bias=bias_root, init=init) : nothing
@@ -1605,8 +1604,7 @@ function (l::TransformerConv)(g::GNNGraph, x::AbstractMatrix,
     end
 
     if l.skip_connection
-        @assert size(h, 1) == size(x, 1) "In-channels must correspond to \
-            out-channels * heads if skip_connection is used"
+        @assert size(h, 1) == size(x, 1) "In-channels must correspond to out-channels * heads if skip_connection is used"
         h += x
     end
     if !isnothing(l.BN1)
--- a/src/layers/conv.jl
+++ b/src/layers/conv.jl
@@ -1465,7 +1465,7 @@ edge features.
 It further contains options to also be configured as the transformer-like multi head 
 attention convolutional operator from the 
 [Attention, Learn to Solve Routing Problems!](https://arxiv.org/abs/1706.03762) paper,
-including a successiv feed-forward network as well as skip layers and batch normalization.
+including a successive feed-forward network as well as skip layers and batch normalization.
 
 The layer's forward pass is given by
 ```math
--- a/src/layers/conv.jl
+++ b/src/layers/conv.jl
@@ -1448,3 +1448,203 @@ function Base.show(io::IO, l::EGNNConv)
     end
     print(io, ")")
 end
+
+
+@doc raw"""
+    TransformerConv((in, ein) => out; heads=1, concat=true, init=glorot_uniform,
+        add_self_loops=false, bias_qkv=false, bias_root=true, root_weight=true, beta=false)
+
+The transformer-like multi head attention convolutional operator from the 
+[Masked Label Prediction: Unified Message Passing Model for Semi-Supervised 
+Classification](https://arxiv.org/abs/2009.03509) paper, which also considers 
+edge features.
+It further contains options to also be configured as the transformer-like multi head 
+attention convolutional operator from the 
+[Attention, Learn to Solve Routing Problems!](https://arxiv.org/abs/1706.03762) paper,
+including a successiv feed-forward network as well as skip layers and batch normalization.
+
+The layer's forward pass is given by
+```math
+x_i' = W_1x_i + \sum_{j\in N(i)} \alpha_{ij} (W_2 x_j + W_6e_{ij})
+```
+where the attention scores are
+```math
+\alpha_{ij} = \mathrm{softmax}\left(\frac{(W_3x_i)^T(W_4x_j+
+W_6e_{ij})}{\sqrt{d}}\right).
+```
+
+Optionally, a combination of the aggregated value with transformed root node features 
+by a gating mechanism via
+```math
+x'_i = \beta_i W_1 x_i + (1 - \beta_i) \underbrace{\left(\sum_{j \in \mathcal{N}(i)}
+\alpha_{i,j} W_2 x_j \right)}_{=m_i}
+```
+with
+```math
+\beta_i = \textrm{sigmoid}(W_5^{\top} [ W_1 x_i, m_i, W_1 x_i - m_i ]).
+```
+can be performed.
+
+# Arguments 
+
+- `in`: Dimension of input features, which also corresponds to the dimension of 
+    the output features.
+- `ein`: Dimension of the edge features; if 0, no edge features will be used.
+- `heads`: Number of heads in output.
+- `concat`: Concatenate layer output or not. If not, layer output is averaged
+    over the heads.
+- `init`: Weight matrices' initializing function.
+- `add_self_loops`: Add self loops to the input graph.
+- `bias_qkv`: If set, bias is used in the key, query and value transformations for nodes.  
+- `bias_root`: If set, the layer will also learn an additive bias for the root when root 
+    weight is used.
+- `root_weight`: If set, the layer will add the transformed root node features
+    to the output.
+- `beta`: If set, will combine aggregation and transformed root node features by
+    gating mechanism.
+- `skip_connection`: If set, a skip connection will be made from the input and 
+    added to the output.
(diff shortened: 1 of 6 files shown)


src/layers/conv.jl
```diff
@@ -1452,3 +1452,200 @@ function Base.show(io::IO, l::EGNNConv)
...
+@functor TransformerConv
+
+Flux.trainable(l::TransformerConv) = (l.W1, l.W2, l.W3, l.W4, l.W5, l.W6, l.FF, l.BN1, l.BN2)
+
+TransformerConv(ch::Pair{Int,Int}, args...; kws...) = TransformerConv((ch[1], 0) => ch[2], args...; kws...)
+
+function TransformerConv(ch::Pair{NTuple{2, Int}, Int}; 
+    heads::Int=1, concat::Bool=true, init=glorot_uniform, add_self_loops::Bool=false, 
+    bias_qkv=true, bias_root::Bool=true, root_weight::Bool=true, beta::Bool=false, 
+    skip_connection::Bool=false, batch_norm::Bool=false, ff_channels::Int=0)
```
CarloLucibello, line 1543:
```suggestion
        heads::Int = 1, 
        concat::Bool = true, 
        init = glorot_uniform, 
        add_self_loops::Bool = false, 
        bias_qkv = true, 
        bias_root::Bool = true, 
        root_weight::Bool = true, 
        beta::Bool = false, 
        skip_connection::Bool = false, 
        batch_norm::Bool = false, 
        ff_channels::Int = 0)
```

graidl:
Note that the Julia 1.6 CI run indicates two failed tests, however not in respect to `TransformerConv` but `GATConv`, which the PR doesn't touch. I tried to reproduce these test failures on my side, but it seems they only happen with a small probability. Maybe the tests are not deterministic, and it would be a good idea to include a constant random seed in testrun.jl and/or the individual test files?

CarloLucibello:
> Maybe the tests are not deterministic

yes, this is the case. Let's see if the recent changes in the numerical stability of GATConv have made them more robust

src/layers/conv.jl
```diff
@@ -1452,3 +1452,200 @@ function Base.show(io::IO, l::EGNNConv)
...
+    if l.concat
+        h = reshape(h, out * heads, :)  # concatenate heads
+    else
+        h = mean(h, dims=2)  # average heads
+        h = reshape(h, out, :)
+    end
+
+    if !isnothing(W1x)  # root_weight
+        if !isnothing(l.W5)  # beta
+            beta = l.W5(vcat(h, W1x, h - W1x))
```
CarloLucibello, line 1599:
```suggestion
            beta = l.W5(vcat(h, W1x, h .- W1x))
```

src/layers/conv.jl
```diff
@@ -1452,3 +1452,200 @@ function Base.show(io::IO, l::EGNNConv)
...
+            h += W1x
+        end
+    end
+
+    if l.skip_connection
+        @assert size(h, 1) == size(x, 1) "In-channels must correspond to out-channels * heads if skip_connection is used"
+        h += x
+    end
+    if !isnothing(l.BN1)
+        h = l.BN1(h)
```
CarloLucibello, line 1611: should this be LayerNorm instead? (eq.6 https://arxiv.org/pdf/2009.03509.pdf)

src/layers/conv.jl
CarloLucibello: also maybe `beta=true` would be a better default so that we match the paper

src/layers/conv.jl
```diff
@@ -1452,3 +1452,200 @@ function Base.show(io::IO, l::EGNNConv)
     end
     print(io, ")")
 end
+
+
+@doc raw"""
+    TransformerConv((in, ein) => out; heads=1, concat=true, init=glorot_uniform,
+        add_self_loops=false, bias_qkv=false, bias_root=true, root_weight=true, beta=false)
```
CarloLucibello, line 1459: `skip_connection`, `batch_norm` and `ff_channels` are missing. 
Since there are so many arguments better to write them here as
```
TransformerConv((in, ein) => out; [heads, concat, init, add_self_loops, bias_qkv, bias_root, root_weight, beta]))
```
and add the defaults to the Arguments section

src/layers/conv.jl
```diff
@@ -1452,3 +1452,200 @@ function Base.show(io::IO, l::EGNNConv)
...
+    over the heads.
+- `init`: Weight matrices' initializing function.
+- `add_self_loops`: Add self loops to the input graph.
+- `bias_qkv`: If set, bias is used in the key, query and value transformations for nodes.  
+- `bias_root`: If set, the layer will also learn an additive bias for the root when root 
+    weight is used.
+- `root_weight`: If set, the layer will add the transformed root node features
+    to the output.
+- `beta`: If set, will combine aggregation and transformed root node features by
+    gating mechanism.
```
CarloLucibello, line 1507: better call this option `gating` instead of `beta`.

CarloLucibello:
I'm worried about the two papers using two different normalizations, maybe we should consider layers just as basic blocks and try to not conflate too much. 

So I suggest the following:
- `TransformerConv`  implements the Masked Label Prediction paper, and from the current state of the PR we remove `skip_connection`, `batch_norm`,  `ff_channels`. So basically TransformerConv here becomes the same as the on in PyG.
- We create a new `models/` folder at the root of this repo, where we try to reproduce the exact architectures from papers. No need for full training examples, just a model constructors. "Kool et al." can be implemented as a GNNChain of TransformerConvs, LayerNorms and MLPs, SkipConnections.

CarloLucibello:
@graidl what do you think? I'm not sure what is the right path here

src/layers/conv.jl
graidl: Thanks for polishing. Concerning the default for beta, I will argue below.

src/layers/conv.jl
graidl: Valid point - I will answer to this below in the general restructuring suggestion.

src/layers/conv.jl
graidl: Thanks for pointing this out, I will take care of this!

src/layers/conv.jl
graidl: Yes, `gating is better!

graidl:
Thanks for reviewing the proposed extension so quickly, your feedback is much appreciated, and you indeed address important questions! Let me start with a little background so that you understand my motivation for some of the so far made decisions: Originally, I started with the model from the newer paper from Kool et al. https://arxiv.org/abs/1803.08475 on which we rely in one of our research projects. As in the famous paper from Vaswani et al. https://arxiv.org/abs/1706.03762, where the classical Transformer architecture (not related to graphs) has been introduced, Kool et al. apply batch normalization and skip layers in conjunction with the message passing as important building block of the whole model. This building block appears in many works relating to transformers in general, I would even say that they are nowadays the standard ingredients making transformers what they are. As we needed in some of our work also the consideration of edge features, I stomped at some point at the Pytorch Geometric implementation of `TransformerConv`, which relies on the much older paper by Shi et al. https://arxiv.org/pdf/2009.03509.pdf, where today's Transformers have not yet been known. Yes, principles are very similar, yet there are smaller differences here and there, for example in the normalization as you pointed out. Moreover,  also this gating mechanism on top of the attention mechanism feels a bit strange in my opinion and is not used in the more recent work of Kool. Ultimately, I ended up in trying to combine aspects of the different approaches and lent a few ideas of Pytorch's `TransformerConv` implementation. 

Now to your specific suggestions:

- Concerning the gating mechanism (beta): This does not appear in the newer works such as Kool et al., and I would therefore tend towards having it switched off as default. Note that also Pytorch's `TransformerConv` has it switched off by default. In Kool et al., they use the inclusive neighborhood (i.e., in essence self loops and self-attention) to deal with the combination of a node's own features with those of the neighbors, which in my opinion is a simpler way to achieve a similar behavior.

- Concerning normalization, skip layers and the feed-forward layers: I am also quite unsure what the best way is to deal with them. In some way, they feel to me as belonging to what a "transformer makes a transformer", just as the gating mechanism is currently included following the Shi et al paper. However, note that a whole transformer model is much more than this, having an encoder and decoder part etc. Thus, it wouldn't feel right to call the combination of the attention layer with BN and skip layer a (whole) "model" and put it into a new model-folder. It is a just small building block, or a kind of "standard extension" to the bare convolution mechanism. To conclude, I would still tend towards leaving BN, skip layers and the FF part as suggested as optional components in `TransformerConv`, but I don't have a strong opinion on this. If you feel it is better to keep things slim, I am equally happy with removing these parts. Clearly, for the users of the library/layer, it should not be a big deal to compose such structures on their own.

Let me know how you prefer to proceed, and I will prepare a revision considering all your remarks, thanks!

CarloLucibello:
I'm merging this PR as it is, beside its utility per se the layer is a very good template for implementing many transformer-like blocks.
Thanks for the effort, an impeccable PR.

PR #245, opened 2022-12-23
update tutorials

CarloLucibello:

PR #250, opened 2022-12-29
Add new cache for changed notebooks!

Dsantra92:
We need to commit the cache notebooks every time we make some changes to the pluto notebooks. This pr refreshes cache files for changes made in #245. We should mention somewhere how to contribute to docs.


CarloLucibello:
Thanks! It would be great if you could add a section about caching tutorials in the Developer Notes of the docs

Dsantra92:
> Thanks! It would be great if you could add a section about caching tutorials in the Developer Notes of the docs

Yes, that's a good idea.

Dsantra92:
docs-usage

Needed for usage in the docs ✌🏾

CarloLucibello:
thanks!

PR #291, opened 2023-05-30
Add TemporalDataStore

aurorarossi:
This PR implements TemporalDataStore.

### DIFF
--- a/src/GNNGraphs/datastore.jl
+++ b/src/GNNGraphs/datastore.jl
@@ -1,9 +1,9 @@
 """
     DataStore([n, data])
-    DataStore([n,] k1 = x1, k2 = x2, ...)
+    DataStore([n], k1 = x1, k2 = x2, ...)
 
 A container for feature arrays. The optional argument `n` enforces that
-`numobs(x) == n` for each array contained in the datastore.
+`numobs(x) == n` for each array contained in the `Datastore``.
 
 At construction time, the `data` can be provided as any iterables of pairs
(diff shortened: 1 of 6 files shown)


src/GNNGraphs/datastore.jl
```diff
@@ -1,9 +1,9 @@
 """
     DataStore([n, data])
-    DataStore([n,] k1 = x1, k2 = x2, ...)
+    DataStore([n], k1 = x1, k2 = x2, ...)
 
 A container for feature arrays. The optional argument `n` enforces that
-`numobs(x) == n` for each array contained in the datastore.
+`numobs(x) == n` for each array contained in the `Datastore``.
```
CarloLucibello, line 6:
```suggestion
`numobs(x) == n` for each array contained in the `DataStore`.
```

CarloLucibello:
A few comments: 
- Any particular reason for having time as the last dimension instead of nodes? Node dimension in the GNN context is similar to the batch dimension. In the Flux ecosystem sequences (e.g. in NLP) have shape `[batch_dim, seq_len, embed_dim]`, so here we probably want to have `[num_nodes, t, embed_dim]` and `[num_edges, t, embed_dim]`.
- Do we need a `TemporalDataStore` structure at all? Can't we just use `DataStore`, since it can store arbitrary dimension arrays already? We could extend it to optionally check the size of the second to last dimension as well. 
- Is it enough to have a data store that assumes that features exist at all times, or do we need something more flexible?

PR #293, opened 2023-06-07
Add TemporalSnapshotsGNNGraph struct

aurorarossi:


CarloLucibello:
Nice! Documentation is missing but can come in a follow up PR

PR #296, opened 2023-06-12
make `add_snapshot` non mutating

CarloLucibello:
... + add some fixes for HeteroGraphs + add some documentation


src/GNNGraphs/temporalsnapshotsgnngraph.jl
```diff
@@ -93,51 +97,88 @@ num_snapshots: 6
 """
 function add_snapshot(tg::TemporalSnapshotsGNNGraph, t::Int, g::GNNGraph)
     @assert g.num_nodes == tg.num_nodes[t] "number of nodes must match"
```
aurorarossi, line 99:
```suggestion
    @assert g.num_nodes == tg.num_nodes[1] "number of nodes must match"
```
 `t` gives error when you want to append a snapshot at the end

PR #313, opened 2023-06-21
Add random temporal hyperbolic graph

aurorarossi:


src/GNNGraphs/generate.jl
```diff
@@ -362,3 +362,96 @@ function rand_temporal_radius_graph(number_nodes::Int,
     end
     return TemporalSnapshotsGNNGraph(tg)
 end
+
+
+function _hyperbolic_distance(nodeA::Array{Float64, 1},nodeB::Array{Float64, 1}; ζ::Real)
+    if nodeA!=nodeB
+        d=acosh(cosh(ζ*nodeA[1])*cosh(ζ*nodeB[1])-(sinh(ζ*nodeA[1])*sinh(ζ*nodeB[1])*cos(pi-abs(pi-abs(nodeA[2]-nodeB[2])))))/ζ
```
CarloLucibello, line 369:
```suggestion
    if nodeA != nodeB
        d=acosh(cosh(ζ*nodeA[1])*cosh(ζ*nodeB[1])-(sinh(ζ*nodeA[1])*sinh(ζ*nodeB[1])*cos(pi-abs(pi-abs(nodeA[2]-nodeB[2])))))/ζ
```
whitespaces around operators missing everywhere and the long ling should be broken into multiple ones for readability

src/GNNGraphs/generate.jl
```diff
@@ -362,3 +362,96 @@ function rand_temporal_radius_graph(number_nodes::Int,
...
+                                   number_snapshots::Int;
+                                   α::Real,
+                                   R::Real,
+                                   speed::Real,
+                                   ζ::Real=1,
+                                   self_loop = false,
+                                   kws...)
+
+Create a random temporal graph given `number_nodes` nodes and `number_snapshots` snapshots.
+First, the positions of the nodes are generated with a quasi-uniform distribution (depending on the parameter `α`) in hyperbolic space within a disk of radius `R`. Two nodes are connected if their hyperbolic distance is less than `R`. Each following snapshot is created in order to keep the same initial distribution.
```
CarloLucibello, line 387: What does quasi-uniform mean? 

Also, "each following snapshot is created in order to keep the same initial distribution." is not clear

aurorarossi: The term quasi-uniform is used in the paper, but it has no formal definition. For a certain value of the parameter alpha the distribution is uniform, for others it is not. In the paper, quasi-uniform refers informally to the values of the parameter that are close to those for which the distribution is uniform.

The positions of the points in each snapshot are updated so that they are distributed with the same distribution as in the first snapshot (the marginal distribution is the same for each snapshot).

PR #314, opened 2023-07-07
Add `TGCNCell` layer

aurorarossi:
This PR adds the layer described in the paper: [T-GCN: A Temporal Graph Convolutional Network for Traffic Prediction](https://arxiv.org/pdf/1811.05320.pdf).

### DIFF
--- a/src/layers/temporalconv.jl
+++ b/src/layers/temporalconv.jl
@@ -15,6 +15,7 @@ struct TGCNCell <: GNNLayer
     dense_update_gate::Dense
     dense_reset_gate::Dense
     dense_candidate_state::Dense
+    in::Int
     out::Int
 end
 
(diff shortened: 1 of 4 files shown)


src/layers/temporalconv.jl
```diff
@@ -0,0 +1,70 @@
+function set_hidden_state(x::AbstractMatrix, h::Union{AbstractMatrix, Nothing}, out::Int)
```
CarloLucibello, line 1: Let's use more specific names, like tgcn_set_hiddenstate

src/layers/temporalconv.jl
```diff
@@ -0,0 +1,70 @@
...
+- `use_edge_weight`: If `true`, consider the edge weights in the input graph (if available).
+                     If `add_self_loops=true` the new weights will be set to 1. 
+                     This option is ignored if the `edge_weight` is explicitly provided in the forward pass.
+                     Default `false`.
+"""
+struct TGCNCell <: GNNLayer
+    conv::GCNConv
+    dense_update_gate::Dense
+    dense_reset_gate::Dense
+    dense_candidate_state::Dense
```
CarloLucibello, line 36: Can we use a GruCell here instead?

src/layers/temporalconv.jl
```diff
@@ -0,0 +1,59 @@
...
+    gru = Flux.GRUv3Cell(out, out)
+    return TGCNCell(conv, gru, in,
+                    out)
+end
+
+function (tgcn::TGCNCell)(g::GNNGraph, x::AbstractArray; h = nothing)
+    h = _tgcn_set_hidden_state(x, h, tgcn.out)
+    x̃ = tgcn.conv(g, x)
+    h, _ = tgcn.gru(h, x̃)
+    return h
```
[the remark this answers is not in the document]
aurorarossi: I also added the `state0` field, which will be useful when implementing `Recur`.

CarloLucibello (approved):

Issue #332, opened 2023-09-03
HeteroGraphConv bug: ERROR: duplicate field name in NamedTuple: "movie" is not unique


svilupp:
First of all, thank you for this package! It's amazing - I've been exploring it for some use cases at work!

I think I noticed a bug in HeteroGraphConv when one adds two destination nodes of the same name.

Expectation: Two destination nodes of the same type in HeteroGraphConv should be aggregated by the `aggr` kwarg

Behavior: When two destination nodes clash (same node type), an error is thrown (see MWE):
> ERROR: duplicate field name in NamedTuple: "movie" is not unique
>
> Stacktrace:
>  [1] NamedTuple
>    @ Core ./boot.jl:620 [inlined]
>  [2] (NamedTuple{(:movie, :movie)})(itr::Vector{Matrix{Float32}})
>    @ Base ./namedtuple.jl:149
>  [3] _reduceby_node_t(aggr::Function, outs::Vector{Matrix{Float32}}, ntypes::Vector{Symbol})
>    @ GraphNeuralNetworks ~/Documents/GitHub/GraphNeuralNetworks.jl/src/layers/heteroconv.jl:80
>  [4] (::HeteroGraphConv)(g::GNNHeteroGraph{Tuple{…}}, x::@NamedTuple{user::Matrix{…}, movie::Matrix{…}, actor::Matrix{…}})
>    @ GraphNeuralNetworks ~/Documents/GitHub/GraphNeuralNetworks.jl/src/layers/heteroconv.jl:65

Suspected root cause: [This line](https://github.com/CarloLucibello/GraphNeuralNetworks.jl/blob/9a4f4b7d86afc707cd52283c229b676915a9f382/src/layers/heteroconv.jl#L79). `_reduceby_node_t` assumes `ntypes` to be a unique list of destination node types, but it is not.

MWE:
```julia
using GraphNeuralNetworks

g = GNNHeteroGraph((:user, :rate, :movie) => ([1,1,2,3], [7,13,5,7]);num_nodes=Dict(:user=>10,:movie=>15))
g = add_edges(g, (:actor, :like, :movie) => ([1,2,3,3,3], [3,5,1,9,4]))

x = (user = rand(Float32, 64, 10), movie = rand(Float32, 64, 15),actor=rand(Float32, 64, 3));
layer = HeteroGraphConv((:user, :rate, :movie) => GraphConv(64 => 32),
[...]

PR #333, opened 2023-09-03
Fix heteroconv aggregation + tests

svilupp:
Fixes https://github.com/CarloLucibello/GraphNeuralNetworks.jl/issues/332

Proposal:
- adds `unique()` to `_reduceby_node_t` to construct a valid tuple
- `unique` operation is blocked from ChainRulesCore
- adds tests for HeteroGraphConv aggregation

Question:
- Are you okay with this implementation?
- Is there a cleaner way to avoid trying to take gradients over `unique`? (I didn't want to block the whole signature, because it's a fairly generic one)

### DIFF
--- a/test/layers/heteroconv.jl
+++ b/test/layers/heteroconv.jl
@@ -33,4 +33,62 @@
                                 (:B, :to, :A) => GraphConv(64 => 32, relu));
         @test length(layer.etypes) == 2
     end
+
+    @testset "Destination node aggregation" begin
+        # deterministic setup to validate the aggregation
+        d, n = 3, 5
+        g = GNNHeteroGraph(((:A, :to, :B) => ([1, 1, 2, 3], [1, 2, 2, 3]),
+                (:B, :to, :A) => ([1, 1, 2, 3], [1, 2, 2, 3]),
+                (:C, :to, :A) => ([1, 1, 2, 3], [1, 2, 2, 3])); num_nodes = Dict(:A => n, :B => n, :C => n))
+        model = HeteroGraphConv([
+                (:A, :to, :B) => GraphConv(d => d, init = ones, bias = false),
+                (:B, :to, :A) => GraphConv(d => d, init = ones, bias = false),
+                (:C, :to, :A) => GraphConv(d => d, init = ones, bias = false)]; aggr = +)
+        x = (A = rand(Float32, d, n), B = rand(Float32, d, n), C = rand(Float32, d, n))
+        y = model(g, x)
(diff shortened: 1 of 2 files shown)


test/layers/heteroconv.jl
```diff
@@ -33,4 +33,62 @@
...
+                (:C, :to, :A) => GraphConv(d => d, init = ones, bias = false)]; aggr = +)
+        x = (A = rand(Float32, d, n), B = rand(Float32, d, n), C = rand(Float32, d, n))
+        y = model(g, x)
+        weights = ones(Float32, d, d)
+
+        ### Test default summation aggregation
+        # B2 has 2 edges from A and itself (sense check)
+        expected = sum(weights * x.A[:, [1, 2]]; dims = 2) .+ weights * x.B[:, [2]]
+        output = y.B[:, [2]]
+        @assert expected ≈ output
```
CarloLucibello, line 55: all these `@assert` should become `@test`

CarloLucibello:
Nice, seems the right thing to do! Also very well tested, thanks.

CarloLucibello:
> Is there a cleaner way to avoid trying to take gradients over unique? (I didn't want to block the whole signature, because it's a fairly generic one)

The current solution is fine and we can merge as it is. At some point `unique` will be fixed by https://github.com/JuliaDiff/ChainRules.jl/pull/736

test/layers/heteroconv.jl
svilupp: ha! Apologies, I was lazy to import Test when developing in my scratch file and then forgot to replace 🤦‍♂️

svilupp:
I've checked the nightly error. 
It's in the same file as I've changed ("test/layers/heteroconv.jl"), but it seems to be unrelated. It passes locally on 1.10 beta2 and 1.9.3.

Error comes from outside the tests in the `gradient()` call, probably because of the getindex access?

`grad, dx = gradient((model, x) -> sum(model(g, x)[1]) + sum(model(g, x)[2].^2), model, x)`

Error:

> HeteroGraphConv: Error During Test at /home/runner/work/GraphNeuralNetworks.jl/GraphNeuralNetworks.jl/test/layers/heteroconv.jl:1
>   Got exception outside of a @test
>   Can't differentiate boundscheck expression $(Expr(:boundscheck)).
>   You might want to check the Zygote limitations documentation.
>   https://fluxml.ai/Zygote.jl/latest/limitations

CarloLucibello:
Flux itself is failing on nightly so at the moment I'm not paying attention to it.

Issue #341, opened 2023-09-14
The constraint in Flux.batch(gs::AbstractVector{<:GNNHeteroGraph}) does not seem to be strong enough


AarSeBail:
While I'd also like there to be no constraint, I do not believe that the current constraint is sufficient.

https://github.com/CarloLucibello/GraphNeuralNetworks.jl/blob/066d8f50c88f4e558be50c9f45b7fbdea8bb6527/src/GNNGraphs/transform.jl#L585C3-L585C3

Consider for example the following:
```
using GraphNeuralNetworks
using Flux

gs = [rand_heterograph((:A =>10, :B => 14), ((:A, :to1, :A) => 5, (:A, :to2, :B) => 20)),
                   rand_heterograph((:A =>10, :B => 14), ((:A, :to1, :A) => 5))]

g = Flux.batch(gs)
```

The code errors with `ERROR: KeyError: key (:A, :to2, :B) not found` due to 

```julia
num_edges = Dict(edge_t => sum(g.num_edges[edge_t] for g in gs) for edge_t in etypes)
```

I believe that the assertion needs to be something along the following lines.

```julia
# TODO remove these constraints
@assert ntypes == intersect([g.ntypes for g in gs]...)
@assert etypes == intersect([g.etypes for g in gs]...)
```

PR #342, opened 2023-09-14
Remove type constraints on Flux.batch for GNNHeteroGraph 

AarSeBail:
I removed the aforementioned type constraints by generalizing the Flux.batch implementation. First and foremost, the implementation may not be correct, despite passing tests, so that needs to be checked.

It's not the cleanest patch so it definitely needs some revisions at the very least. In particular, instead of edge_index I use a nullable version, and the same with get_edge_weight. Would it be worth exposing these functions? Also, I am sure that there are inefficiencies in the patch.

The test case is incomplete, so it needs to be fleshed out correctly.

Resolves https://github.com/CarloLucibello/GraphNeuralNetworks.jl/issues/341

### DIFF
--- a/src/GNNGraphs/datastore.jl
+++ b/src/GNNGraphs/datastore.jl
@@ -125,7 +125,6 @@ function Base.setproperty!(ds::DataStore, s::Symbol, x)
 end
 
 Base.getindex(ds::DataStore, s::Symbol) = getproperty(ds, s)
-Base.setindex!(ds::DataStore, s::Symbol, x) = setproperty!(ds, s, x)
 Base.setindex!(ds::DataStore, x, s::Symbol) = setproperty!(ds, s, x)
 
 function Base.show(io::IO, ds::DataStore)
--- a/src/GNNGraphs/datastore.jl
+++ b/src/GNNGraphs/datastore.jl
@@ -126,6 +126,7 @@ end
 
 Base.getindex(ds::DataStore, s::Symbol) = getproperty(ds, s)
 Base.setindex!(ds::DataStore, s::Symbol, x) = setproperty!(ds, s, x)
+Base.setindex!(ds::DataStore, x, s::Symbol) = setproperty!(ds, s, x)
 
 function Base.show(io::IO, ds::DataStore)
     len = length(ds)
--- a/src/GNNGraphs/transform.jl
+++ b/src/GNNGraphs/transform.jl
@@ -580,7 +580,11 @@ end
 
 function Flux.batch(gs::AbstractVector{<:GNNHeteroGraph})
     function edge_index_nullable(g::GNNHeteroGraph{<:COO_T}, edge_t::EType)
-        get(g.graph, edge_t, (nothing, nothing))[1:2]
+        if haskey(g.graph, edge_t)
+            g.graph[edge_t][1:2]
+        else
+            nothing
+        end
     end
 
     function get_edge_weight_nullable(g::GNNHeteroGraph{<:COO_T}, edge_t::EType)
@@ -594,7 +598,7 @@ function Flux.batch(gs::AbstractVector{<:GNNHeteroGraph})
     v_num_nodes = Dict(node_t => [get(g.num_nodes, node_t, 0) for g in gs] for node_t in ntypes)
     num_nodes = Dict(node_t => sum(v_num_nodes[node_t]) for node_t in ntypes)
     num_edges = Dict(edge_t => sum(get(g.num_edges, edge_t, 0) for g in gs) for edge_t in etypes)
-    edge_indices = Dict(edge_t => filter(x -> x[1] !== nothing, [edge_index_nullable(g, edge_t) for g in gs]) for edge_t in etypes)
+    edge_indices = edge_indices = Dict(edge_t => [edge_index_nullable(g, edge_t) for g in gs] for edge_t in etypes)
     nodesum = Dict(node_t => cumsum([0; v_num_nodes[node_t]])[1:(end - 1)] for node_t in ntypes)
     graphs = []
     for edge_t in etypes
@@ -604,8 +608,8 @@ function Flux.batch(gs::AbstractVector{<:GNNHeteroGraph})
         #     @show ei[1]
         # end 
         # # [ei[1] for (ii, ei) in enumerate(edge_indices[edge_t])]
-        s = cat_features([ei[1] .+ nodesum[src_t][ii] for (ii, ei) in enumerate(edge_indices[edge_t])])
-        t = cat_features([ei[2] .+ nodesum[dst_t][ii] for (ii, ei) in enumerate(edge_indices[edge_t])])
+        s = cat_features([ei[1] .+ nodesum[src_t][ii] for (ii, ei) in enumerate(edge_indices[edge_t]) if ei !== nothing])
+        t = cat_features([ei[2] .+ nodesum[dst_t][ii] for (ii, ei) in enumerate(edge_indices[edge_t]) if ei !== nothing])
         w = cat_features(filter(x -> x !== nothing, [get_edge_weight_nullable(g, edge_t) for g in gs]))
         push!(graphs, edge_t => (s, t, w))
     end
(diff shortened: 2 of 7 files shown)


src/GNNGraphs/transform.jl
```diff
@@ -579,17 +579,22 @@ end
 
 
 function Flux.batch(gs::AbstractVector{<:GNNHeteroGraph})
+    function edge_index_nullable(g::GNNHeteroGraph{<:COO_T}, edge_t::EType)
+        get(g.graph, edge_t, (nothing, nothing))[1:2]
```
CarloLucibello, line 583:
```suggestion
        if haskey(g.graph, edge_t)
            get(g.graph, edge_t, (nothing, nothing))[1:2]
        else
            nothing
        end
```

src/GNNGraphs/transform.jl
```diff
@@ -579,17 +579,22 @@ end
...
-    @assert ntypes == gs[1].ntypes
-    @assert etypes == gs[1].etypes
     
-    v_num_nodes = Dict(node_t => [get(g.num_nodes,node_t,0) for g in gs] for node_t in ntypes)
+    v_num_nodes = Dict(node_t => [get(g.num_nodes, node_t, 0) for g in gs] for node_t in ntypes)
     num_nodes = Dict(node_t => sum(v_num_nodes[node_t]) for node_t in ntypes)
-    num_edges = Dict(edge_t => sum(g.num_edges[edge_t] for g in gs) for edge_t in etypes)
-    edge_indices = Dict(edge_t => [edge_index(g, edge_t) for g in gs] for edge_t in etypes)
+    num_edges = Dict(edge_t => sum(get(g.num_edges, edge_t, 0) for g in gs) for edge_t in etypes)
+    edge_indices = Dict(edge_t => filter(x -> x[1] !== nothing, [edge_index_nullable(g, edge_t) for g in gs]) for edge_t in etypes)
```
CarloLucibello, line 597:
```suggestion
    edge_indices = Dict(edge_t => [edge_index_nullable(g, edge_t) for g in gs] for edge_t in etypes)
```

src/GNNGraphs/transform.jl
```diff
@@ -601,32 +606,40 @@ function Flux.batch(gs::AbstractVector{<:GNNHeteroGraph})
...
+    s = edge_index(gs[1], gs[1].etypes[1])[1] # grab any source vector
 
-    function materialize_graph_indicator(g, node_t)     
-        ones_like(s, g.num_nodes[node_t])
+    function materialize_graph_indicator(g, node_t)
+        x = get(g.num_nodes, node_t, nothing)
+        if x !== nothing
+            x = ones_like(s, x)
+        end
+        x
```
CarloLucibello, line 624:
```suggestion
        n = get(g.num_nodes, node_t, 0)
        return ones_like(s, n)
```

src/GNNGraphs/transform.jl
```diff
@@ -601,32 +606,40 @@ function Flux.batch(gs::AbstractVector{<:GNNHeteroGraph})
...
     end
-    v_gi = Dict(node_t => [materialize_graph_indicator(g, node_t) for g in gs] for node_t in ntypes)
+    v_gi = Dict(node_t => filter(x -> x !== nothing, [materialize_graph_indicator(g, node_t) for g in gs]) for node_t in ntypes)
```
CarloLucibello, line 626:
```suggestion
    v_gi = Dict(node_t => [materialize_graph_indicator(g, node_t) for g in gs] for node_t in ntypes)
```

src/GNNGraphs/transform.jl
```diff
@@ -601,32 +606,40 @@ function Flux.batch(gs::AbstractVector{<:GNNHeteroGraph})
         # # [ei[1] for (ii, ei) in enumerate(edge_indices[edge_t])]
         s = cat_features([ei[1] .+ nodesum[src_t][ii] for (ii, ei) in enumerate(edge_indices[edge_t])])
         t = cat_features([ei[2] .+ nodesum[dst_t][ii] for (ii, ei) in enumerate(edge_indices[edge_t])])
```
CarloLucibello, line 608:
```suggestion
        s = cat_features([ei[1] .+ nodesum[src_t][ii] for (ii, ei) in enumerate(edge_indices[edge_t]) if ei !== nothing])
        t = cat_features([ei[2] .+ nodesum[dst_t][ii] for (ii, ei) in enumerate(edge_indices[edge_t]) if ei !== nothing])
```
It seems to me that the previous version was wrong since `edge_idices[edge_t]` could be of length lengths less than `length(nodesum[src_t])`

CarloLucibello:
Thanks for taking this, it seems mostly fine, added a few suggestions. 
> Would it be worth exposing these functions? 

I don't think they are worth exposing at the moment

AarSeBail:
I was expanding the test case (I'm glad I did, I believe I uncovered an issue with cat_features), and I came across something that seems off. The syntax `g.ndata[:nodetype][:x] = value` seems broken, unless I'm mistaken. I think the issue is that in datastore.jl, Base.setindex! is defined as ```julia
Base.setindex!(ds::DataStore, s::Symbol, x) = setproperty!(ds, s, x)```

I think the order is incorrect, and should be
```julia
Base.setindex!(ds::DataStore, x, s::Symbol) = setproperty!(ds, s, x)```

Is this intentional and I'm just using it incorrectly?

**Comment by AarSeBail** [comment]
I believe that everything here is in working order now. I modified `cat_features(xs::AbstractVector{<:Dict})` to be less strict, and the current tests do still pass. I am wondering whether it would be worthwhile updating the other cat_features implementations which might pose problems for heterogeneous graphs, such as `cat_features(x1::Dict{Symbol, T}, x2::Dict{Symbol, T})`. In particular the ones which throw `@error "cannot concatenate feature data with different keys"`. If so, I would not be opposed to drafting a PR.

I kept the previous implementation of `Base.setindex!(ds::DataStore, s::Symbol, x)`, but added an implementation of `Base.setindex!(ds::DataStore, x, s::Symbol)` to make the syntax work as I believe it was intended.

I also finalized the new test case.

File: src/GNNGraphs/datastore.jl [hunk 6]
```diff
@@ -126,6 +126,7 @@ end
 
 Base.getindex(ds::DataStore, s::Symbol) = getproperty(ds, s)
 Base.setindex!(ds::DataStore, s::Symbol, x) = setproperty!(ds, s, x)
```
CarloLucibello, line 128:
```suggestion
```

src/GNNGraphs/utils.jl
```diff
@@ -119,13 +119,11 @@ end
...
     _allkeys = [sort(collect(keys(x))) for x in xs]
-    _keys = _allkeys[1]
-    all(y -> y == _keys, _allkeys) ||
-        @error "cannot concatenate feature data with different keys"
+    _keys = union(_allkeys...)
     length(xs) == 1 && return xs[1]
 
     # concatenate 
-    return Dict([k => cat_features([x[k] for x in xs]) for k in _keys]...)
+    return Dict([k => cat_features([x[k] for x in filter(x->haskey(x, k),xs)]) for k in _keys]...)
```
CarloLucibello, line 126: This seems more readable:
```suggestion
    return Dict([k => cat_features([x[k] for x in xs if haskey(x, k)]) for k in _keys]...)
```

CarloLucibello:
i can't believe that `setindex!` was not covered by tests! Can you add an explicit one in `test/GNNGraphs/datastore.jl`?

> I am wondering whether it would be worthwhile updating the other cat_features implementations which might pose problems for heterogeneous graphs, such as cat_features(x1::Dict{Symbol, T}, x2::Dict{Symbol, T}). In particular the ones which throw @error "cannot concatenate feature data with different keys". If so, I would not be opposed to drafting a PR.

it could be worth doing

AarSeBail:
I have added the test that you requested and committed the suggestions. I'm not sure why it did not pass the test on CI earlier, since it was passing locally. Assuming that it is working now, I think this is near completion.

test/GNNGraphs/datastore.jl
```diff
@@ -16,7 +16,7 @@ end
 @testset "getproperty / setproperty!" begin
     x = rand(10)
     ds = DataStore(10, (:x => x, :y => rand(2, 10)))
-    @test ds.x == ds[:x] == x
+    @test ds.x == ds[:x] == x # ds[:x] tests getindex!, not setindex!, which is called by ds[:x] = v
```
CarloLucibello, line 19:
```suggestion
    @test ds.x == ds[:x] == x
```

PR #345, opened 2023-09-26
Add heterogeneous add_self_loop support

AarSeBail:


AarSeBail:
Regarding the method `get_edge_weight_nullable`, if it keeps popping up in heterogeneous graphs, it may be worth refining.

src/GNNGraphs/transform.jl
```diff
@@ -38,6 +38,62 @@ function add_self_loops(g::GNNGraph{<:ADJMAT_T})
...
+        get(g.graph, edge_t, (nothing, nothing, nothing))[3]
+    end
+
+    src_t, _, tgt_t = edge_t
+    (src_t === tgt_t) ||
+        @error "cannot add a self-loop with different source and target types"
+    
+    n = get(g.num_nodes, src_t, 0)
+
+    # By avoiding using haskey, this only calls ht_keyindex once instead of twice
```
CarloLucibello, line 62:
```suggestion
```
These kinds of performance concerns for dictionary queries are irrelevant. The heavy operations are the copies and the concatenations. Therefore the concern here should be to make the code has readable as possible, which means using haskey in my opinion.

src/GNNGraphs/transform.jl
```diff
@@ -38,6 +38,62 @@ function add_self_loops(g::GNNGraph{<:ADJMAT_T})
...
+    if (x = get(g.graph, edge_t, nothing)) !== nothing
+        s, t = x[1:2]
+        nodes = convert(typeof(s), [1:n;])
+        s = [s; nodes]
+        t = [t; nodes]
+    else
+        nodes = [1:n;]
```
CarloLucibello, line 69: This creates a cpu array even if the rest of graph lives on gpu. We should create a new array similar to one in the existing relations.

AarSeBail:
I believe this to now be initializing edge arrays with the correct type, but please check to confirm. I also extended the test cases marginally.

Tests should pass on latest, but not on nightly, similarly to #356.

CarloLucibello:
nice, thanks!

PR #347, opened 2023-09-27
example correction

### LINKED ISSUE #346: Error in Example Code
The example code at [https://carlolucibello.github.io/GraphNeuralNetworks.jl/stable/](https://carlolucibello.github.io/GraphNeuralNetworks.jl/stable/) has a small problem:

```julia
[...]
for epoch in 1:100
    [...]
    @info (; epoch, train_loss=loss(train_loader), test_loss=loss(test_loader))
end
```

There is no function signature for `loss(dl::DataLoader)`.  The fix is obviously simple: add the required `model` arg--PR inbound 😉

I could also make a PR to implement CI testing on the docs examples, if that would be welcome.  I like doing that for my packages, since this kind of thing has a way of popping up.

eahenle:
fixes #346 by adding the missing `model` argument to calls to `loss`

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -82,6 +82,6 @@ for epoch in 1:100
         Flux.update!(opt, model, grad[1])
     end
 
-    @info (; epoch, train_loss=loss(train_loader), test_loss=loss(test_loader))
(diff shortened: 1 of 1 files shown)


CarloLucibello:
whoops, thanks!

PR #355, opened 2023-12-28
First draft GPS conv layer

### LINKED ISSUE #351: GPSConv
Would be fun to have the GPSConv layer in here and I'd be happy to take a crack at it (if deemed useful).

Should be straight forward, except the "global attention" part. Not sure if there is already an implementation somewhere to just add as a dependency (transformers.jl as dependency might be a bit too much?) or if implementing it here directly would be the way to go. any thoughts?

https://arxiv.org/abs/2205.12454

https://pytorch-geometric.readthedocs.io/en/latest/generated/torch_geometric.nn.conv.GPSConv.html#:~:text=The%20GPS%20layer%20is%20based,operates%20on%20the%20input%20graph.

abieler:
This is only first "mock" version of a [GPSConv](https://arxiv.org/abs/2205.12454) layer to see if we would want it in the Repo in that form.

- Adds a `DotProductAttention` layer that uses `NNlib.dot_product_attention()`
- Adds a `GPSConv` layer
  - has the `DotPRoductAttention` as global attention layer
  - takes a conv-layer as local message passing
- Not sure about the `GNNChain()` implementation, if it should stay where it is or move into the struct?

- JuliaFormatter() got a bit too greedy and made some changes here and there, I can revert those of course
- Did not check for correctness of the implementation yet

Let me know what you think and I can adjust / keep going from here.

Close #351

### DIFF
--- a/src/layers/conv.jl
+++ b/src/layers/conv.jl
@@ -74,33 +74,31 @@ end
 @functor GCNConv
 
 function GCNConv(ch::Pair{Int, Int}, σ = identity;
-                 init = glorot_uniform,
-                 bias::Bool = true,
-                 add_self_loops = true,
-                 use_edge_weight = false)
+        init = glorot_uniform,
+        bias::Bool = true,
+        add_self_loops = true,
+        use_edge_weight = false)
     in, out = ch
     W = init(out, in)
     b = bias ? Flux.create_bias(W, true, out) : false
     GCNConv(W, b, σ, add_self_loops, use_edge_weight)
 end
 
-check_gcnconv_input(g::GNNGraph{<:ADJMAT_T}, edge_weight::AbstractVector) = 
+function check_gcnconv_input(g::GNNGraph{<:ADJMAT_T}, edge_weight::AbstractVector)
     throw(ArgumentError("Providing external edge_weight is not yet supported for adjacency matrix graphs"))
+end
 
 function check_gcnconv_input(g::GNNGraph, edge_weight::AbstractVector)
-    if length(edge_weight) !== g.num_edges 
+    if length(edge_weight) !== g.num_edges
         throw(ArgumentError("Wrong number of edge weights (expected $(g.num_edges) but given $(length(edge_weight)))"))
     end
 end
 
 check_gcnconv_input(g::GNNGraph, edge_weight::Nothing) = nothing
 
-
-function (l::GCNConv)(g::GNNGraph, 
-                      x::AbstractMatrix{T},
-                      edge_weight::EW = nothing
-                      ) where {T, EW <: Union{Nothing, AbstractVector}}
-
+function (l::GCNConv)(g::GNNGraph,
+        x::AbstractMatrix{T},
+        edge_weight::EW = nothing) where {T, EW <: Union{Nothing, AbstractVector}}
     check_gcnconv_input(g, edge_weight)
 
     if l.add_self_loops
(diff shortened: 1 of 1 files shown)


src/layers/conv.jl
```diff
@@ -74,33 +74,31 @@ end
 @functor GCNConv
 
 function GCNConv(ch::Pair{Int, Int}, σ = identity;
-                 init = glorot_uniform,
-                 bias::Bool = true,
-                 add_self_loops = true,
-                 use_edge_weight = false)
```
CarloLucibello, line 80: this PR is doing a lot of unrelated changes, please remove them

src/layers/conv.jl
```diff
@@ -223,6 +221,60 @@ function Base.show(io::IO, l::ChebConv)
...
+    attnlayer::GNNChain
+    ffn::GNNChain
+end
+
+function GPSConv(ch::Pair{Int, Int}, gconv, σ = identity; init = glorot_uniform)
+    in, out = ch
+    gattn = DotProductAttention(ch)
+    convlayer = GNNChain(gconv, Dropout(0.5), LayerNorm(out))
+    attnlayer = GNNChain(gattn, Dropout(0.5), LayerNorm(out))
+    ffn = GNNChain(Dense(in => 2 * in, σ), Dropout(0.5), Dense(2 * in => in), Dropout(0.5))
```
CarloLucibello, line 262: does the paper specify that we need dropout in the MLP?

src/layers/conv.jl
```diff
@@ -223,6 +221,60 @@ function Base.show(io::IO, l::ChebConv)
...
+
+function (l::GPSConv)(g::AbstractGNNGraph, x)
+    check_num_nodes(g, x)
+    f = GNNChain(
+            Parallel(+,
+                Parallel(+, l.convlayer, identity),
+                Parallel(+, l.attnlayer, identity),
+            ),
+            l.ffn
+    )
```
CarloLucibello, line 274: We should avoid constructing a gnnchain in each forward pass. Also, the use of parallel makes the sequence of operations hard to parse. 
We should just have a plain sequence of transformations of the input here, like:
```julia
x1 = l.bn1(l.dropout1(l.conv(g, x)) .+ x))
x2 =l.bn2(l.dropout2(l.attn(x)) .+ x))
y = l.mlp( x1 + x2) # not sure if we should also add a skip connection here
```
Notice the order of operations. It is different from what the layer is currently doing. Current implementation doesn't seem to follow eqs.  (6)-(11) in the paper.

CarloLucibello:
Thanks, this is a nice start. A few comments:

- No need to introduce the `DotProductAttention` type, we can use the [MultiHeadAttention](https://fluxml.ai/Flux.jl/stable/models/layers/#Flux.MultiHeadAttention) from Flux. According to the table A.2-A.5 in the paper, multi-head attention is the preferred choice. We should have an `nheads` argument in the constructor. See also https://pytorch-geometric.readthedocs.io/en/latest/generated/torch_geometric.nn.conv.GPSConv.html for the kind of flexibility we could try to achieve (in several PRs). 

- Part of the contribution of the paper is the discussion of different types of embeddings. This package lacks many of these embeddings. I hope they will be added in the future but in any case, it is ok for this PR to only implement the layer.

- I think the current order of operations is wrong, see comment https://github.com/CarloLucibello/GraphNeuralNetworks.jl/pull/355#discussion_r1438471782

- BathcNorm should be used instead of LayerNorm

- In the paper is not clear if we should apply a residual connection after the MLP. For figure D.1 it seems there is one, but there is none according to Eq. 11.

abieler:
Thanks for the comments. I'll be going over the `authors codebase` > `paper` > `pytorch` implementation for implementation details for the next version

i.e.
- they support batch- and layer-norm in their [codbase](https://github.com/rampasek/GraphGPS/blob/main/graphgps/layer/gps_layer.py#L129)
- do have a [skip connection](https://github.com/rampasek/GraphGPS/blob/main/graphgps/layer/gps_layer.py#L225) after last ff layer

Issue #277, opened 2023-04-27
GCNConv without normalization


FDemelas:
Hello, I'm trying to use GCNConv with negative edge weights, hence I would like to deactivate the normalization as clearly the square root of negative value is problematic.
Is it possible in some way to deactivate the normalization (but still use the edge weights) or change with a personalized normalization?

CarloLucibello:
Hi, we don't support that currently. It would be nice to add the possibility of a custom normalization (PR welcome!). 
As a quick workaround you can define your own custom layer by copying and pasting the GCNConv implementation and 
removing the normalization
https://github.com/CarloLucibello/GraphNeuralNetworks.jl/blob/a7de26fc65e46f6f29208b8d365b9c0e79a1390a/src/layers/conv.jl#L125-L126

FDemelas:
Thank you very much

rbSparky:
Seems like a quick fix

@CarloLucibello should I create a PR that adds the possibility of custom normalization in GCNConv?

PR #356, opened 2024-01-01
Custom Norm Func for GCNConv

rbSparky:
Fixes issue [Issue #277](https://github.com/CarloLucibello/GraphNeuralNetworks.jl/issues/277)

- Added custom norm fn
- Changed tests

### DIFF
--- a/src/layers/conv.jl
+++ b/src/layers/conv.jl
@@ -38,6 +38,9 @@ Takes as input a graph `g`,ca node feature matrix `x` of size `[in, num_nodes]`,
 and optionally an edge weight vector. Returns a node feature matrix of size 
 `[out, num_nodes]`.
 
+The `norm_fn` parameter allows for custom normalization of the graph convolution operation by passing a function as argument. 
+By default, it computes ``\frac{1}{\sqrt{d}}`` i.e the inverse square root of the degree (`d`) of each node in the graph. 
+
 # Examples
 
 ```julia
--- a/src/layers/conv.jl
+++ b/src/layers/conv.jl
@@ -100,7 +100,7 @@ check_gcnconv_input(g::GNNGraph, edge_weight::Nothing) = nothing
 function (l::GCNConv)(g::GNNGraph, 
                       x::AbstractMatrix{T},
                       edge_weight::EW = nothing,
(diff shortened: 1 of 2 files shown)


src/layers/conv.jl
```diff
@@ -32,7 +32,7 @@ and optionally an edge weight vector.
 
 # Forward
 
-    (::GCNConv)(g::GNNGraph, x::AbstractMatrix, edge_weight = nothing) -> AbstractMatrix
+    (::GCNConv)(g::GNNGraph, x::AbstractMatrix, edge_weight = nothing, normalization_fn::Function = (d) -> 1 ./ sqrt.(d)) -> AbstractMatrix
```
CarloLucibello, line 35:
```suggestion
    (::GCNConv)(g::GNNGraph, x::AbstractMatrix, edge_weight = nothing, norm_fn::Function = d -> 1 ./ sqrt.(d)) -> AbstractMatrix
```
I prefer a shorter name.

src/layers/conv.jl
```diff
@@ -98,7 +99,8 @@ check_gcnconv_input(g::GNNGraph, edge_weight::Nothing) = nothing
 
 function (l::GCNConv)(g::GNNGraph, 
                       x::AbstractMatrix{T},
-                      edge_weight::EW = nothing
+                      edge_weight::EW = nothing,
+                      normalization_fn::Function = (d) -> 1 ./ sqrt.(d)
```
CarloLucibello, line 103:
```suggestion
                      norm_fn::Function = d -> 1 ./ sqrt.(d)  
```

CarloLucibello:
Thanks, this looks nice. The docstring should better specify how the norm function is used though.

rbSparky:
> Thanks, this looks nice. The docstring should better specify how the norm function is used though.

Done

CarloLucibello:
tests are not passing

rbSparky:
> tests are not passing

Seems like in the ``Julia nightly`` release, there is some issue that seems to be unrelated to the change:

**ERROR**: The following 2 direct dependencies failed to precompile:
cuDNN [02a925ec-e4fe-4b08-9a7e-0d78e3d38ccd]

I suppose that it would be enough to get the ``Julia 1 - ubuntu-latest`` tests to pass?

Issue #311, opened 2023-06-20
Layers support for HeteroGraphConv


CarloLucibello:
HeteroGraphConv are build out of standard graph conv layers which are individually applied to the different relations. 
The list of layers supporting integration with HeteroGraphConv should be extended. 

- [ ] `AGNNConv`        
- [ ] `CGConv`         
- [ ] `ChebConv`       
- [ ] `EGNNConv`       
- [ ] `EdgeConv`       
- [ ] `GATConv`        
- [ ] `GATv2Conv`      
- [ ] `GatedGraphConv` 
- [ ] `GCNConv`        | 
- [ ] `GINConv`        
- [ ] `GMMConv`        
- [x] `GraphConv`
- [ ] `MEGNetConv`    
- [ ] `NNConv`        
- [ ] `ResGatedGraphConv`
- [ ] `SAGEConv`       
- [ ] `SGConv`         
- [ ] `TransformerConv`

codetalker7:
Hi @CarloLucibello. Is this a documentation issue? I looked at the implementation in `src/layers/heteroconv.jl`, and it seems like the implementation takes care of all standard graph conv layers (in the sense that, the `layers` field in `HeteroGraphConv` can be any `GNNLayer`, and the code should work for any layer type). Am I missing something?

CarloLucibello:
No it is actually an implementation issue that involves a tiny change to the forward pass of each layer. In the orginal PR (#300) it was done only for `GraphConv`.
One needs to relax the forward signature to take an `AbstractGNNGraph` instead of a `GNNGraph` and then insert at the beginning the line 
```julia
xj, xi = expand_srcdst(g, x)
```
since during the hetero message passing `x` is a tuple containing the src and dst features (relative to two different node types).

codetalker7:
Hi @CarloLucibello, another small question: for layers like `AGNNConv` which involve self loops; such layers will then have to operate on two types of edges, right? Because, we'll have to add self-loops for nodes of the target type (this is being done currently in the `AGNNConv` implementation: line 1018 of https://github.com/CarloLucibello/GraphNeuralNetworks.jl/blob/master/src/layers/conv.jl).

Also, motivated by this: it seems like adding self loops to `GNNHeteroGraph`s is not supported. Will it be a good idea to add this functionality, wherein we can add self-loops for a particular node type?

CarloLucibello:
[...]

PR #367, opened 2024-01-28
feat: Add GCNConv support for HeteroGraphConv

askorupka:
This PR adds GCNConv support for HeteroGraphConv.
This is related to issue https://github.com/CarloLucibello/GraphNeuralNetworks.jl/issues/311.


src/layers/conv.jl
```diff
@@ -118,29 +127,33 @@ function (l::GCNConv)(g::GNNGraph,
...
     Dout, Din = size(l.weight)
-    if Dout < Din
-        # multiply before convolution if it is more convenient, otherwise multiply after
-        x = l.weight * x
-    end
-    if edge_weight !== nothing
-        d = degree(g, T; dir = :in, edge_weight)
+
+    if g isa GNNHeteroGraph
+        # not the best practice but this loop is
```
askorupka, line 132: @CarloLucibello I need your guidance how to change that part so that it can use correct edge type as an input instead of this dummy solution. The problem is that edge type is not passed to the GCNConv layer (only to HeteroConv) so I'm not sure how to add it.

src/layers/conv.jl
```diff
@@ -118,29 +127,33 @@ function (l::GCNConv)(g::GNNGraph,
...
-        d = degree(g, T; dir = :in, edge_weight)
+
+    if g isa GNNHeteroGraph
+        # not the best practice but this loop is 
+        # a temporary mechanism to insure matching dimensions 
+        # until it'll be possible to pass edge here
+        for e in g.etypes |> collect
+            d = degree(g, e; dir = :in)
+            length(d) == size(xi[2]) ? break : continue 
+        end
```
askorupka, line 138: @CarloLucibello I need your guidance how to change that part so that it can use correct edge type as an input instead of this dummy solution. The problem is that edge type is not passed to the GCNConv layer (only to HeteroConv) so I'm not sure how to add it.

PS
This issue will repeat itself in several more layers needing HeteroGraphConv support.

src/layers/conv.jl
```diff
@@ -118,29 +121,27 @@ function (l::GCNConv)(g::GNNGraph,
         end
     end
     Dout, Din = size(l.weight)
-    if Dout < Din
-        # multiply before convolution if it is more convenient, otherwise multiply after
-        x = l.weight * x
-    end
```
CarloLucibello, line 124: Why this has been removed?

src/layers/conv.jl
```diff
@@ -118,29 +121,27 @@ function (l::GCNConv)(g::GNNGraph,
...
-    if edge_weight !== nothing
-        d = degree(g, T; dir = :in, edge_weight)
+
+    if g isa GNNHeteroGraph
+        d = degree(g, edge_t; dir = :in)
     else
-        d = degree(g, T; dir = :in, edge_weight = l.use_edge_weight)
+        if edge_weight !== nothing
+            d = degree(g; dir = :in, edge_weight)
```
CarloLucibello, line 129: Degree should receive an eltype `T` as well. Probably the CI failure is due to `d` here having eltype `Int`
Then `1 ./ sqrt.(d)` is a float64 vector and this propagates to the final output

src/layers/conv.jl
askorupka: if the multiplication happens before convolution, I used to have an error about dimensions or `MethodError: no method matching *(::Matrix{Float32}, ::Tuple{Matrix{Float32}, Matrix{Float32}})`
as the comment says `if it is more convenient`, I understood these lines are not a must and cut them out

src/layers/conv.jl
askorupka: T is not defined now even in the function signature, as we need to remove that constraint for Heterograph.
Also, default for T is nothing, and when `T = nothing` is passed to the degree function, type should be inferred from the type of `src`. I think it should be inferred from the feature matrix instead?

src/layers/conv.jl
CarloLucibello: We do want that performance optimization

src/layers/conv.jl
CarloLucibello: You can define T = eltype(xj)

askorupka: I know, just did it 7 minutes ago, waiting for tests :wheel_of_dharma:

src/layers/conv.jl
askorupka: it's either this optimization and `DimensionMismatch: arrays could not be broadcast to a common size; got a dimension with lengths 4 and 2` error message (so no support for GCNConv layer) or no optimization and no error message.
if there is third option, I'm willing to implement it, but need a hint what's that. I tried to get rid of that error for long already

askorupka:
@CarloLucibello ready for re-review. code is programatically working and passing tests.
needed to introduce conditional behavior in order for both tests to pass (one for GCNConv regular, one for GCNConv for GNNHeteroGraph)

Project.toml
```diff
@@ -12,6 +12,7 @@ Functors = "d9f16b24-f501-4c13-a1f2-28368ffc5196"
 Graphs = "86223c79-3864-5bf0-83f7-82e725a168b6"
 KrylovKit = "0b1a1467-8014-51b9-945f-bf0ae24f4b77"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
+MLDatasets = "eb30cadb-4394-5ae3-aed4-317e484a6458"
```
CarloLucibello, line 15:
```suggestion
```

src/layers/conv.jl
```diff
@@ -118,26 +121,40 @@ function (l::GCNConv)(g::GNNGraph,
...
-    end
+        x = x .* c' # xj should be same but x = xj .* c' test for conv.jl failing
+        if edge_weight !== nothing
+            x = propagate(e_mul_xj, g, +, xj = x, e = edge_weight)
+        elseif l.use_edge_weight
+            x = propagate(w_mul_xj, g, +, xj = x)
+        else
+            x = propagate(copy_xj, g, +, xj = x)
+        end
+    end
```
CarloLucibello, line 155: there is too much code duplication here, we should try to collapse these two branches

src/layers/conv.jl
```diff
@@ -118,26 +121,40 @@ function (l::GCNConv)(g::GNNGraph,
...
+            d = degree(g, T; dir = :in, edge_weight = l.use_edge_weight)
+        end
     end
     c = norm_fn(d)
-    x = x .* c'
-    if edge_weight !== nothing
-        x = propagate(e_mul_xj, g, +, xj = x, e = edge_weight)
-    elseif l.use_edge_weight
-        x = propagate(w_mul_xj, g, +, xj = x)
+    if xi != xj
```
CarloLucibello, line 137:
```suggestion
    if xi !== xj
```
we should check identity, not equality

PR #369, opened 2024-01-29
Add `TemporalGraphConv`

aurorarossi:
This PR adds a `TemporalGraphConv` constructor to each snapshot of the `TemporalSnapshotsGNNGraph`, a convolutional homogeneous graph `layer`. 
A PR will follow to update the documentation.

### DIFF
--- a/src/layers/temporalconv.jl
+++ b/src/layers/temporalconv.jl
@@ -187,7 +187,6 @@ function Base.show(io::IO, a3tgcn::A3TGCN)
     print(io, "A3TGCN($(a3tgcn.in) => $(a3tgcn.out))")
 end
 
-
 @doc raw"""
     TemporalGraphConv(layer)
 
--- a/src/layers/temporalconv.jl
+++ b/src/layers/temporalconv.jl
@@ -186,3 +186,25 @@ end
 function Base.show(io::IO, a3tgcn::A3TGCN)
     print(io, "A3TGCN($(a3tgcn.in) => $(a3tgcn.out))")
 end
+
+
+# struct to apply convolutional layers to each snapshot of the temporal graph
+
+struct TemporalGraphConv
+    layer::GNNLayer
+end
+
+Flux.@functor TemporalGraphConv
+
+function (tgconv::TemporalGraphConv)(tsg::TemporalSnapshotsGNNGraph)
+    return TemporalSnapshotsGNNGraph(
+        tsg.num_nodes,
+        tsg.num_edges,
+        tsg.num_snapshots,
+        tgconv.layer.(tsg.snapshots),
+        tsg.tgdata)
+end
+
+function (tgconv::TemporalGraphConv)(tsg::TemporalSnapshotsGNNGraph, x)
+        return tgconv.layer.(tsg.snapshots, x)
+end
--- a/src/layers/temporalconv.jl
+++ b/src/layers/temporalconv.jl
@@ -206,7 +206,6 @@ julia> size(temp_gcn(tsg).ndata.x[1])
 (5, 20)
 ```
 """
-
 struct TemporalGraphConv
     layer::GNNLayer
 end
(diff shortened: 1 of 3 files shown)


src/layers/temporalconv.jl
```diff
@@ -186,3 +186,42 @@ end
...
+julia> tsg = TemporalSnapshotsGNNGraph(snapshots);
+
+julia> temp_gcn = TemporalGraphConv(GCNConv(3 => 5, relu))
+TemporalGraphConv(GCNConv(3 => 5, relu))
+
+julia> size(temp_gcn(tsg).ndata.x[1])
+(5, 20)
+```
+"""
+
```
CarloLucibello, line 209:
```suggestion
```

src/layers/temporalconv.jl
```diff
@@ -186,3 +186,42 @@ end
 function Base.show(io::IO, a3tgcn::A3TGCN)
     print(io, "A3TGCN($(a3tgcn.in) => $(a3tgcn.out))")
 end
+
+@doc raw"""
+    TemporalGraphConv(layer)
+
+A constructor for applying to each snapshot of the `TemporalSnapshotsGNNGraph` a convolutional homogeneous graph `layer`.
+
```
CarloLucibello, line 194: Paper citation is missing, and also more explanation of how it works. 

Am I correct in saying that this applies `layer` independently to each temporal slice?

src/layers/temporalconv.jl
```diff
@@ -186,3 +186,42 @@ end
...
+julia> temp_gcn = TemporalGraphConv(GCNConv(3 => 5, relu))
+TemporalGraphConv(GCNConv(3 => 5, relu))
+
+julia> size(temp_gcn(tsg).ndata.x[1])
+(5, 20)
+```
+"""
+
+struct TemporalGraphConv
+    layer::GNNLayer
```
CarloLucibello, line 211:
```suggestion
struct TemporalGraphConv{L <: GNNLayer}
    layer::L
```

src/layers/temporalconv.jl
```diff
@@ -186,3 +186,42 @@ end
...
+    return TemporalSnapshotsGNNGraph(
+        tsg.num_nodes,
+        tsg.num_edges,
+        tsg.num_snapshots,
+        tgconv.layer.(tsg.snapshots),
+        tsg.tgdata)
+end
+
+function (tgconv::TemporalGraphConv)(tsg::TemporalSnapshotsGNNGraph, x)
+        return tgconv.layer.(tsg.snapshots, x)
```
CarloLucibello, line 226:
```suggestion
    return tgconv.layer.(tsg.snapshots, x)
```

Also, here broadcasting should be avoided, since both `snapshots` and `x` could be expanded

src/layers/temporalconv.jl
CarloLucibello: >  I would say it is similar to the HeteroGraphConv for heterographs.

Yes, although HeteroGraphConv in the end aggregates messages coming from different edge types, while here there is no cross-talk between different times, it has to happen in separate modules (e.g. temporal attention in the last reference above)

src/layers/temporalconv.jl
CarloLucibello: sorry, I got confused, it is the same

CarloLucibello:
I think that instead of introducing a new layer, we can extend all layers to take a temporalsnapshotgraph as input and apply itself timewise. The time dimension in this way is like a batch dimension. 
It would be something like
```julia
function (l::GNNLayer)(g::TemporalSnapshotsGNNGraph, x::AbstractVector; kws...)
    y = ....
    return y
end  
```
Maybe the operation could be made more efficient batching the snapshots first and then unbatching the output y later, but this is a performance optimization that can be explored later.

aurorarossi:
Ah ok I see. Do you prefer one big PR with support for TemporalSnapshotsGNNGraphs for all layers or one PR for each layer?
In both cases I will close this PR.
Thanks!

PR #371, opened 2024-01-29
feat: support weights when generating from SimpleWeightedGraph

### LINKED ISSUE #85: Weights not included in GNNGraph made from SimpleWeightedDiGraph
Hello, just found your sweet package! 
I ran into a minor issue when wrapping a SimpleWeightedDiGraph with a GNNGraph. The weights from the DiGraph are not included in the wrapped graph.

Below is an MWE of this behavior. I'm not certain if this was by design.
```
using Graphs, GraphNeuralNetworks, SimpleWeightedGraphs
function randGraph(graphSize::Int)
       graph = rand(graphSize, graphSize)
       foreach(enumerate(eachcol(graph))) do (idx, col)
             graph[idx, :] .= col
             graph[idx, idx] = 0
       end
       return graph
end
a = randGraph(10)
g = SimpleWeightedDiGraph(a)
b = GNNGraph(g)
b.graph
```
RETURNS
```
([1, 1, 1, 1, 1, 1, 1, 1, 1, 2  …  9, 10, 10, 10, 10, 10, 10, 10, 10, 10], [2, 3, 4, 5, 6, 7, 8, 9, 10, 1  …  10, 1, 2, 3, 4, 5, 6, 7, 8, 9], nothing)
```

askorupka:
This PR add support for weights when generating from SimpleWeightedGraph.
This fixes #85.

```julia
julia> srcs = [1, 2, 1]
julia> dsts = [2, 3, 3]
julia> wts = [0.5, 0.8, 2.0]
julia> g = SimpleWeightedGraph(srcs, dsts, wts)
julia> gnn_g = GNNGraph(g)

julia> get_edge_weight(gnn_g)
6-element Vector{Float64}:
 0.5
 2.0
 0.5
 0.8
 2.0
 0.8
```


src/GNNGraphs/GNNGraphs.jl
```diff
@@ -16,6 +16,7 @@ using LinearAlgebra, Random, Statistics
 import MLUtils
 using MLUtils: getobs, numobs, ones_like, zeros_like
 import Functors
+using SimpleWeightedGraphs
```
CarloLucibello, line 19: We don't want to carry an extra dependency, so this should be part of an [extension](https://pkgdocs.julialang.org/v1/creating-packages/#Conditional-loading-of-code-in-packages-(Extensions)).

Also, the compact entry is missing.

askorupka:
@CarloLucibello ready for re-review.
Needed to remove test as this required putting SimpleWeightedGraphs as a dependency, which is contradictory to the point of moving them to the extension.

ext/GraphNeuralNetworksSimpleWeightedGraphsExt/GNNGraphs/gnngraph.jl
```diff
@@ -0,0 +1,23 @@
+__precompile__(false)
```
CarloLucibello, line 1:
```suggestion

```

ext/GraphNeuralNetworksSimpleWeightedGraphsExt/GNNGraphs/gnngraph.jl
```diff
@@ -0,0 +1,23 @@
...
+
+function GraphNeuralNetworks.GNNGraph(g::T; edge_weight = nothing, kws...) where 
+                 {T <: Union{SimpleWeightedGraph, SimpleWeightedDiGraph, 
+                       GraphNeuralNetworks.AbstractGraph}}
+    s = Graphs.src.(Graphs.edges(g))
+    t = Graphs.dst.(Graphs.edges(g))
+    if g isa Union{SimpleWeightedGraph, SimpleWeightedDiGraph}
+        w = filter(!iszero, g.weights |> vec) |> collect
```
CarloLucibello, line 9: What is the need of the filtering? And why the collect?

CarloLucibello: Also, if edge_weight != nothing we should use that instead

ext/GraphNeuralNetworksSimpleWeightedGraphsExt/GNNGraphs/gnngraph.jl
```diff
@@ -0,0 +1,23 @@
...
+    end
+    if !Graphs.is_directed(g)
+        # add reverse edges since GNNGraph is directed
+        s, t = [s; t], [t; s]
+        if !isnothing(w) & !(g isa Union{SimpleWeightedGraph, SimpleWeightedDiGraph})
+            @assert length(w) == Graphs.ne(g) "edge_weight must have length equal to the number of undirected edges"
+            w = [w; w]
+        end
+    end
+    num_nodes::Int = Graphs.nv(g)
```
CarloLucibello, line 21:
```suggestion
    num_node = Graphs.nv(g)
```

ext/GraphNeuralNetworksSimpleWeightedGraphsExt/GNNGraphs/gnngraph.jl
```diff
@@ -0,0 +1,23 @@
...
+    GraphNeuralNetworks.GNNGraph((s, t, w); num_nodes = num_nodes, kws...)
```
CarloLucibello, line 22:
```suggestion
    return GNNGraph((s, t, w); num_nodes, kws...)
```

ext/GraphNeuralNetworksSimpleWeightedGraphsExt/GraphNeuralNetworksSimpleWeightedGraphsExt.jl
```diff
@@ -0,0 +1,10 @@
+module GraphNeuralNetworksSimpleWeightedGraphsExt
+
+using GraphNeuralNetworks
+using Graphs
+import GraphNeuralNetworks: AbstractGraph
+using SimpleWeightedGraphs
+
+include("GNNGraphs/gnngraph.jl")
```
CarloLucibello, line 8: let's just put everything here, no need to have a second file, I don't expect this extension to grow much

CarloLucibello:
> Needed to remove test as this required putting SimpleWeightedGraphs as a dependency, which is contradictory to the point of moving them to the extension.

SimpleWeightedGraphs can be a test dependency, you have to put it under [extras] in the Project.toml.
The test can live in `test/ext/GraphNeuralNetworksSimpleWeightedGraphsExt/GraphNeuralNetworksSimpleWeightedGraphsExt.jl`

ext/GraphNeuralNetworksSimpleWeightedGraphsExt/GNNGraphs/gnngraph.jl
askorupka: I'm afraid it will yield the precompiling warning about overwriting the method from GNN package.

ext/GraphNeuralNetworksSimpleWeightedGraphsExt/GNNGraphs/gnngraph.jl
askorupka: Filtering is due to the format of SimpleWeightedGraph `g.weights` which include zero entries for not connected nodes (unlike GNN weights). 

Collect is to cast values to type expected by GNN graph weights.

`Also, if edge_weight != nothing we should use that instead`
I think no, this implementation is specific for SimpleWeightedGraphs weights. I assume creation of GNN graph with weights is already supported and this extension is aiming at adressing SimpleWeightedGraphs conversion specifically. I would leave other cases out of it, not to lengthen developer feedback loop when not necessary (I mean loading another package).

ext/GraphNeuralNetworksSimpleWeightedGraphsExt/GNNGraphs/gnngraph.jl
askorupka: update: not only warning but
```julia
WARNING: Method definition (::Type{GraphNeuralNetworks.GNNGraphs.GNNGraph{T} where T<:Union{AbstractArray{T, 2} where T, Tuple{T, T, V} where V where T<:(AbstractArray{var"#s4", 1} where var"#s4"<:Integer)}})(T) where {T<:(Graphs.AbstractGraph{T} where T)} in module GNNGraphs at /home/agata/.julia/dev/GraphNeuralNetworks/src/GNNGraphs/gnngraph.jl:173 overwritten in module GraphNeuralNetworksSimpleWeightedGraphsExt at /home/agata/.julia/dev/GraphNeuralNetworks/ext/GraphNeuralNetworksSimpleWeightedGraphsExt/GNNGraphs/gnngraph.jl:3.
ERROR: Method overwriting is not permitted during Module precompilation. Use `__precompile__(false)` to opt-out of precompilation
```

ext/GraphNeuralNetworksSimpleWeightedGraphsExt/GNNGraphs/gnngraph.jl
askorupka:
```julia
julia> g.weights
3×3 SparseArrays.SparseMatrixCSC{Float64, Int64} with 6 stored entries:
  ⋅   0.5  2.0
 0.5   ⋅   0.8
 2.0  0.8   ⋅ 

julia> w = filter(!iszero, g.weights |> vec)
6-element SparseArrays.SparseVector{Float64, Int64} with 6 stored entries:
       ⋮
  [6]  =  0.8

julia> w = filter(!iszero, g.weights |> vec) |> collect
6-element Vector{Float64}:
 0.5
 2.0
 0.5
 0.8
 2.0
 0.8
```

ext/GraphNeuralNetworksSimpleWeightedGraphsExt/GNNGraphs/gnngraph.jl
askorupka: There is missing `s` in `num_nodes` in the suggestion, will commit it separately

Project.toml
```diff
@@ -62,6 +65,7 @@ MLDatasets = "eb30cadb-4394-5ae3-aed4-317e484a6458"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 Zygote = "e88e6eb3-aa80-5325-afca-941959d7151f"
 cuDNN = "02a925ec-e4fe-4b08-9a7e-0d78e3d38ccd"
+SimpleWeightedGraphs = "47aef6b3-ad0c-573a-a1e2-d07658019622"
```
CarloLucibello, line 68: These are in alphabetic order

askorupka:
@CarloLucibello may I ask for help what else is required to get rid of the error message as below during tests:
```
Error: Error during loading of extension GraphNeuralNetworksSimpleWeightedGraphsExt of GraphNeuralNetworks, use `Base.retry_load_extensions()` to retry.
│   exception =
│    1-element ExceptionStack:
│    ArgumentError: Package GraphNeuralNetworksSimpleWeightedGraphsExt [df50fc6b-8916-5c00-b5b6-a97a04cc2331] is required but does not seem to be installed:
│     - Run `Pkg.instantiate()` to install all recorded dependencies.
```

PR #373, opened 2024-01-30
refinement: Self loops for HeteroGraph returns g instead of error if src != tgt

askorupka:
This PR changes self loops for HeteroGraph to return g instead of throwing error if src != tgt, as in reality in that case graph doesn't change. This refinement would allow us avoid trycatches or similar mechanism when dealing with conv layers for heterographs.
Also, this unblocks https://github.com/CarloLucibello/GraphNeuralNetworks.jl/pull/367


CarloLucibello:
this should be documented and tested

askorupka:
@CarloLucibello ready for your :eyes:

PR #376, opened 2024-02-14
Adapt `GINConv` to `TemporalSnapshotsGNNGraphs`

aurorarossi:
As discussed in PR #369 , I adapted `GINConv`.
I tried different things with batching but could not get it to work in gradient computation and training.
Something like:

```jl
function (l::GINConv)(tg::TemporalSnapshotsGNNGraph, x::AbstractVector)
   tg = MLUtils.batch(tg.snapshots)
   x = reduce(hcat,x)
   x = l(tg, x)
   x = reshape(x,:, Int.(tg.num_nodes/tg.num_graphs), tg.num_graphs)
   return MLUtils.unbatch(x)
end

function (l::GINConv)(tg::TemporalSnapshotsGNNGraph)
   tg = MLUtils.batch(tg.snapshots)
   tg = l(tg)
   return MLUtils.unbatch(tg) #in this case unbatch was not working on GPU
end
```
If it is ok, a PR will follow to update the documentation.

### DIFF
--- a/src/layers/temporalconv.jl
+++ b/src/layers/temporalconv.jl
@@ -186,3 +186,7 @@ end
 function Base.show(io::IO, a3tgcn::A3TGCN)
     print(io, "A3TGCN($(a3tgcn.in) => $(a3tgcn.out))")
 end
+
+function (l::GINConv)(tg::TemporalSnapshotsGNNGraph, x::AbstractVector)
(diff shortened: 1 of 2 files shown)


CarloLucibello:
It would be great if you could produce some MWE for `batch` and `unbatch` and report the issue upstream to MLUtils.jl

PR #383, opened 2024-02-20
Add SGConv support for HeteroGraphConv 

rbSparky:
Covers Issue #311


rbSparky:
Some tests fail, will fix

CarloLucibello:
this is taking code from #399, so @askorupka should be duly credited. I can add her to the commit message when this will be merged. It is not ready to go though, the code is still broken.

rbSparky:
Will fix and add @askorupka to next commit!

PR #386, opened 2024-02-22
Add EGNNConv support for HeteroGraphConv 

rbSparky:
Covers Issue #311 

This is a work in progress, just wanted to make sure I am on the right track

Since `EGNNConv` has `H` as input as well I added another function:
```
function (hgc::HeteroGraphConv)(g::GNNHeteroGraph, x::NamedTuple, h::AbstractMatrix)
    function forw(l, et)
        sg = edge_type_subgraph(g, et)
        node1_t, _, node2_t = et

        x_features = (x[node1_t], x[node2_t])
        h_features = h # temporary

        return l(sg, h_features, x_features)
    end
    outs = [forw(l, et) for (l, et) in zip(hgc.layers, hgc.etypes)]
    dst_ntypes = [et[3] for et in hgc.etypes]
    return _reduceby_node_t(hgc.aggr, outs, dst_ntypes)
end
```

Let me know if there is an alternative like using the arg in the old function (pass as a `Dict`) but this just seemed more convenient. 

Will add more updates and test in the coming days. **Will remove all debug statements when done.**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ Manifest.toml
 LocalPreferences.toml
 .DS_Store
 /test.jl
+try.jl
--- a/src/layers/conv.jl
+++ b/src/layers/conv.jl
@@ -801,11 +801,10 @@ function SAGEConv(ch::Pair{Int, Int}, σ = identity; aggr = mean,
     SAGEConv(W, b, σ, aggr)
 end
 
-function (l::SAGEConv)(g::AbstractGNNGraph, x)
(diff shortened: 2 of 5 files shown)


CarloLucibello:
is there an idea or some effort planned in fixing this? Otherwise better close so that it won't prevent other people from working on it

rbSparky:
Yes, currently I'm trying to make it work. 
I'll close it if there is no progress

rbSparky:
Could you confirm if I'm on the right path? 
Thanks

PR #387, opened 2024-02-22
Add GINConv support to HeteroGraphConv

rbSparky:
Covers Issue #311 

**Need some advice on how to integrate this in the tests, please let me know.**
For now I have only written a dummy test since I wasn't sure how to test `GINConv` the right way.


CarloLucibello:
you can add a simple test of the output shape, as for the other tests in that file. Then this seems good to go.

PR #390, opened 2024-03-02
Add GINConv support to HeteroGraphConv

rbSparky:
Covers Issue #311 

This operation needs to be done differently for hetero graphs
```
    if g isa GNNHeteroGraph
        #
    else
        l.nn((1 + ofeltype(x, l.ϵ)) * x + m)
    end
```

[EDIT: done]

### DIFF
--- a/src/layers/conv.jl
+++ b/src/layers/conv.jl
@@ -682,7 +682,7 @@ GINConv(nn, ϵ; aggr = +) = GINConv(nn, ϵ, aggr)
 
 function (l::GINConv)(g::AbstractGNNGraph, x)
     check_num_nodes(g, x)
-    xj, _ = expand_srcdst(g, x)
+    xj, _ = expand_srcdst(g, x) # hetero graphs
 
     m = propagate(copy_xj, g, l.aggr, xj = xj)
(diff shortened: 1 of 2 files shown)


rbSparky:
ready for merge?

src/layers/conv.jl
```diff
@@ -680,10 +680,17 @@ Flux.trainable(l::GINConv) = (nn = l.nn,)
...
-    l.nn((1 + ofeltype(x, l.ϵ)) * x + m)
+    xj, xi = expand_srcdst(g, x) 
+ 
+    m = propagate(copy_xj, g, l.aggr, xj = xj)
+    
+    if g isa GNNHeteroGraph
+        l.nn((1 .+ ofeltype(xi, l.ϵ)) .* xi .+ m)
+    else
+        l.nn((1 + ofeltype(x, l.ϵ)) * x + m)
+    end
```
CarloLucibello, line 693: why there is an `if` here instead of just using `xi`?

rbSparky: fixed

rbSparky:
Changes done

PR #397, opened 2024-03-03
Add example for gender classification on `TemporalBrains` dataset

aurorarossi:
This PR adds an example on temporal graph classification, specifically gender classification of 1000 subject graphs extracted from fMRI signals. The example was presented in a [poster](https://hal.science/hal-04230797/document) of mine at the [Julia and Optimization Days 2023](https://julia-users-paris.github.io/workshop/en/index.html).

I will add this example as a tutorial to show the differences between working with graphs where the topology changes over time and those where the topology is fixed but the features are temporal (see tutorial on traffic).


CarloLucibello:
Nice! Maybe you can add a few extra line of comments at the top to give some more context. I agree this would be better as a tutorial.

CarloLucibello (approved):

PR #402, opened 2024-03-05
New `add_self_loops(g)` for hetero graphs

rbSparky:
As mentioned in PR #400, adding a new `add_self_loops(g)` function for hetero graphs would be helpful.

This PR does that. 

Will integrate this function to be used in `conv.jl`.
- [x] SGConv
- [x] GATConv (when #400 is merged)

### DIFF
--- a/src/GNNGraphs/transform.jl
+++ b/src/GNNGraphs/transform.jl
@@ -40,19 +40,23 @@ end
 
 """
     add_self_loops(g::GNNHeteroGraph, edge_t::EType)
+    add_self_loops(g::GNNHeteroGraph)
 
-If the source node type is the same as destination node type in `edge_t`,
-return a graph with the same features as `g` but also adding self-loops 
-of the specified type, `edge_t`. Otherwise it returns `g` unchanged.
+- For a specific edge type (`edge_t`):
+    If the source node type is the same as the destination node type in `edge_t`, return a graph with the same features as `g` but also adding self-loops of the specified type, `edge_t`. Otherwise, it returns `g` unchanged.
+    
+    Nodes with already existing self-loops of type `edge_t` will obtain a second self-loop of type `edge_t`.
+    
+    If the graph has edge weights for edges of type `edge_t`, the new edges will have weight 1.
+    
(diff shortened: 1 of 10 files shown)


CarloLucibello:
No need to reimplement everything. The new method `add_self_loops(hg)` should just call the `add_self_loops(hg, edge_t)` on each edge type. 

Then the docstring here 
https://github.com/CarloLucibello/GraphNeuralNetworks.jl/blob/5250c4dae953c70e6feae27f44df7d7f6a64eacc/src/GNNGraphs/transform.jl#L56
should be describe the additional behavior.

rbSparky:
> No need to reimplement everything. The new method `add_self_loops(hg)` should just call the `add_self_loops(hg, edge_t)` on each edge type.
> 
> Then the docstring here
> 
> https://github.com/CarloLucibello/GraphNeuralNetworks.jl/blob/5250c4dae953c70e6feae27f44df7d7f6a64eacc/src/GNNGraphs/transform.jl#L56
> 
> should be describe the additional behavior.

Changed function and changed docstring to include the new function and what it does

src/GNNGraphs/transform.jl
```diff
@@ -40,18 +40,19 @@ end
...
 
-If the source node type is the same as the destination node type in `edge_t`,
-return a graph with the same features as `g` but also add self-loops 
-of the specified type, `edge_t`. Otherwise, it returns `g` unchanged.
-
-Nodes with already existing self-loops of type `edge_t` will obtain 
-a second set of self-loops of the same type.
-
-If the graph has edge weights for edges of type `edge_t`, the new edges will have weight 1.
+- For a specific edge type (`edge_t`):
```
CarloLucibello, line 45: Let's not start a docstring with a bullet point. 
The change to the docstring can be minimal, just adding a paragraph at the end saying "If `edge_t` is not passed as argument, ..."

rbSparky: Done

rbSparky:
Once this is merged I can make a PR for GATv2 hetero using this function

PR #415, opened 2024-03-09
fix: SGConv layer fixes after implementing support for GNNHeteroGraph

askorupka:
This PR introduces some optimizations and fixes for SGConv.

Fix: 
- previously the loop for propagation involved `x = x .* c'` and it was the only place where `x` was referred to after expansion to `xi` and `xj`. in the propagation process `xj` was passed which made this line of code unnecessary and potentially problematic (remember that for GNNHeteroGraph `x` contains both `xi` and `xj`)

Optimizations:
- reduces number of ifs as definition for `edge_t` not necessary anymore (after changing `add_self_loop` behavior it was used only once in `degree`)
- introduces optimization for multiplying weights before convolution if `Dout < Din` (but only when `g` is not `GNNHeteroGraph`

### DIFF
--- a/src/layers/conv.jl
+++ b/src/layers/conv.jl
@@ -1519,10 +1519,6 @@ function (l::SGConv)(g::AbstractGNNGraph, x,
                      {EW <: Union{Nothing, AbstractVector}}
     @assert !(g isa GNNGraph{<:ADJMAT_T} && edge_weight !== nothing) "Providing external edge_weight is not yet supported for adjacency matrix graphs"
 
-    xj, xi = expand_srcdst(g, x)
-    edge_t = g isa GNNHeteroGraph ? g.etypes[1] : nothing
-    T = eltype(xi)
-
     if edge_weight !== nothing
         @assert length(edge_weight)==g.num_edges "Wrong number of edge weights (expected $(g.num_edges) but given $(length(edge_weight)))"
     end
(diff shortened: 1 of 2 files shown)


src/layers/conv.jl
```diff
@@ -1546,7 +1551,7 @@ function (l::SGConv)(g::AbstractGNNGraph, x,
     end
     c = 1 ./ sqrt.(d)
     for iter in 1:(l.k)
-        x = x .* c'
+        !(g isa GNNHeteroGraph) ? xj = xj .* c' : Nothing
```
CarloLucibello, line 1554: this stuff is not correct, it's not like since it is not a heterograph we can skip normalization. The same goes for gcnconv. 
I think that the right thing would be to compute two normalization constants
```
cin = degree(g, dir=:in)
cout =  degree(g, dir=:out)
```
and use `cout` for the first normalization and `cin` for the second.

src/layers/conv.jl
```diff
@@ -1546,7 +1551,7 @@ function (l::SGConv)(g::AbstractGNNGraph, x,
     end
     c = 1 ./ sqrt.(d)
     for iter in 1:(l.k)
```
CarloLucibello, line 1553: I think that the fact that this layer iterates multiple times makes it unsuitable for heterographs operation. This PR should be closed, I'm sorry, but this operation just doesn't make sense for heterographs.

askorupka: @CarloLucibello does it mean we want to remove GNNHeteroGraph support from SGConv?
if yes, I can do that (and it's okay, it happens).
if no, in my opinion this PR still makes sense with some optimizations. 

I also noticed an oversight: the line 1338 should be amended from `x = x .* c'` to something that accounts for `x` being something else now. I thought I did it as per GCN, my mistake here.

src/layers/conv.jl
askorupka: I was wondering about it, something didn't feel right here. thanks, I'll amend GCN.

askorupka:
closing as per above discussion with @CarloLucibello, SGConv layer GNNHeteroGraph support to be removed by me.

PR #417, opened 2024-03-10
fix: GCNConv support for GNNHeteroGraph

askorupka:
This PR fixes GCNConv layer support for GNNHeteroGraph as per discussion here https://github.com/CarloLucibello/GraphNeuralNetworks.jl/pull/415.

### DIFF
--- a/src/layers/conv.jl
+++ b/src/layers/conv.jl
@@ -127,16 +127,18 @@ function (l::GCNConv)(g::AbstractGNNGraph,
     T = eltype(xi)
 
     if g isa GNNHeteroGraph
-        d = degree(g, g.etypes[1], T; dir = :in)
+        din = degree(g, g.etypes[1], T; dir = :in)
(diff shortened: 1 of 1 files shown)


src/layers/conv.jl
```diff
@@ -127,24 +127,26 @@ function (l::GCNConv)(g::AbstractGNNGraph,
...
+            din = dout = degree(g, T; dir = :in, edge_weight)
         else
-            d = degree(g, T; dir = :in, edge_weight = l.use_edge_weight)
+            din = dout = degree(g, T; dir = :in, edge_weight = l.use_edge_weight)
         end
     end
-    c = norm_fn(d)
-    !(g isa GNNHeteroGraph) ? xj = xj .* c' : Nothing
+    cout = norm_fn(dout)
+    cin = norm_fn(din)
```
CarloLucibello, line 140: we don't want to call norm_fn twice if not needed, so it's better to define `cin` and `cout` in the branches above

PR #424, opened 2024-03-17
Fixes and more tests in `remove_nodes` function

rbSparky:
I tested a little more and there are some errors to be fixed, I will fix those here
I missed these in the last PR

I'll also add more tests to make sure all cases are covered

### DIFF
--- a/src/GNNGraphs/transform.jl
+++ b/src/GNNGraphs/transform.jl
@@ -223,7 +223,7 @@ function remove_nodes(g::GNNGraph{<:COO_T}, nodes_to_remove::AbstractVector)
     function find_edges_to_remove(nodes, nodes_to_remove)
         return findall(node_id -> begin
             idx = searchsortedlast(nodes_to_remove, node_id)
-            idx >= 1 && idx <= length(nodes_to_remove) && nodes_to_remove[idx] == node_id
+            idx >= 1 && idx <= length(nodes_to_remove)
         end, nodes)
     end
     
--- a/src/GNNGraphs/transform.jl
+++ b/src/GNNGraphs/transform.jl
@@ -223,7 +223,7 @@ function remove_nodes(g::GNNGraph{<:COO_T}, nodes_to_remove::AbstractVector)
     function find_edges_to_remove(nodes, nodes_to_remove)
         return findall(node_id -> begin
             idx = searchsortedlast(nodes_to_remove, node_id)
-            idx >= 1 && idx <= length(nodes_to_remove)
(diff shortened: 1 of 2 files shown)


rbSparky:
needed to add these checks to check bounds
```julia
    function find_edges_to_remove(nodes, nodes_to_remove)
        return findall(node_id -> begin
            idx = searchsortedlast(nodes_to_remove, node_id)
            idx >= 1 && idx <= length(nodes_to_remove) && nodes_to_remove[idx] == node_id
        end, nodes)
    end
```

CarloLucibello:
You can add a test that was failing before this PR

rbSparky:
> You can add a test that was failing before this PR

Done, and made the fixes

src/GNNGraphs/transform.jl
```diff
@@ -219,9 +219,16 @@ function remove_nodes(g::GNNGraph{<:COO_T}, nodes_to_remove::AbstractVector)
...
     edata = g.edata
     ndata = g.ndata
-
-    edges_to_remove_s = findall(x -> nodes_to_remove[searchsortedlast(nodes_to_remove, x)] == x, s)
-    edges_to_remove_t = findall(x -> nodes_to_remove[searchsortedlast(nodes_to_remove, x)] == x, t)
+    
+    function find_edges_to_remove(nodes, nodes_to_remove)
+        return findall(node_id -> begin
+            idx = searchsortedlast(nodes_to_remove, node_id)
+            idx >= 1 && idx <= length(nodes_to_remove) && nodes_to_remove[idx] == node_id
```
CarloLucibello, line 226:
```suggestion
            idx >= 1 && idx <= length(nodes_to_remove)
```

rbSparky: This change results in tests failing

rbSparky: I have re added it to see if tests pass

CarloLucibello: why is that? if the idx returned from `searchsortedlast` is in the range then nodes_to_remove[idx] == node_id should be true

rbSparky: `searchsortedlast` doesn't guarantee `nodes_to_remove[i] == node_id` since it only guarantees that node_id is not less than `nodes_to_remove[i]`. 

The condition is necessary to ensure that `node_id` is actually present in `nodes_to_remove`, rather than just being greater than or equal to the last element smaller than `node_id` in the sorted array.

I think this leads to false positives and thus the test failures.

Issue #412, opened 2024-03-07
implementation of the DropMessage paper


rbSparky:
Recently I had planned to work on some experiments related to the [DropMessage paper](https://arxiv.org/abs/2204.10037) and its extensions and application on various datasets like
- Cora
- CiteSeer
- Pubmed
- Flickr
- ogbn-arxiv

If it would be considered useful, I could write up a tutorial/blog related to the work to show GNN.jl users how to modify architectures and use them on various datasets and experiment along the way. 
Experiments include:
- Edge perturbation
- Feature noise injection
- Subgraph sampling
- Diffusion

Let me know if this will be useful for the repository! 😄 
Thanks

rbSparky:
I will use this space to add notes regarding this problem

-  Dropout performs random dropping operation on the node feature matrix, while DropEdge and DropNode, as the name implies, respectively act on the adjacency matrix (edges) and nodes

rbSparky:
After further discussion(on Slack), since it has been decided that it would be more valuable to implement more relevant applications, this discussion thread can now be used for writing the tutorial for "Molecular Property Prediction using EGNNs".

Dataset to be used: QM9

The above techniques from DropMessage can be incorporated as data augmentation technique (to show the user how to write custom functions for tasks like these)

This tutorial is in progress!

PR #427, opened 2024-03-18
Added Personalized PageRank Diffusion [`ppr_diffusion` function]

rbSparky:
Continues Issue #412, takes inspiration from [DGL implementation](https://docs.dgl.ai/en/0.8.x/generated/dgl.transforms.PPR.html)

This PR introduces a new function, `ppr_diffusion`. It calculates Personalized PageRank (PPR) diffusion based on the edge weight matrix and updates the graph with new edge weights derived from the PPR matrix.

Implementation:

The function operates in several key steps:

- Construct Modified Adjacency Matrix (A): Utilizes existing edge weights (w) to build a matrix $A$ representing the graph's structure, adjusted by $(α - 1)$, where $α$ is the damping factor. This adjustment incorporates both direct connections and global graph structure.

```julia

A[dst, src] += w[idx]
A .*= (alpha_f32 - 1)
```
- Incorporate Identity Matrix (I): Adds the identity matrix to A to account for self-transitions, forming $I + (α - 1) \times A$.

```julia

A[i, i] += 1
```
- Apply PPR Formula: Calculates the PPR diffusion matrix as $α \times (I + (α - 1) \times A)^{-1}$, which effectively models the probability of transitioning from one node to another, considering random jumps governed by α.

```julia

PPR = alpha_f32 * inv(A)
```
- Update Edge Weights: Assigns new weights to each edge based on the corresponding values in the PPR matrix, effectively updating the graph's edge weights to reflect the calculated PPR diffusion.

```julia

new_w = [PPR[dst, src] for (src, dst) in zip(s, t)]
```

### DIFF
--- a/src/GNNGraphs/transform.jl
+++ b/src/GNNGraphs/transform.jl
@@ -1098,7 +1098,7 @@ The function performs the following steps:
 # Returns
 - A new `GNNGraph` instance with the same structure as `g` but with updated edge weights according to the PPR diffusion calculation.
 """
-function ppr_diffusion(g::GNNGraph{<:COO_T}; alpha_f32::Float32=0.85f0)
+function ppr_diffusion(g::GNNGraph{<:COO_T}; alpha = 0.85f0)
     s, t = edge_index(g)
     w = ones(Float32, g.num_edges)
 
--- a/src/GNNGraphs/transform.jl
+++ b/src/GNNGraphs/transform.jl
@@ -1079,7 +1079,7 @@ ci2t(ci::AbstractVector{<:CartesianIndex}, dims) = ntuple(i -> map(x -> x[i], ci
 @non_differentiable dense_zeros_like(x...)
 
 """
-    ppr_diffusion(g::GNNGraph{<:COO_T}; alpha_f32::Float32=0.85f0) -> GNNGraph
+    ppr_diffusion(g::GNNGraph{<:COO_T}, alpha =0.85f0) -> GNNGraph
 
 Calculates the Personalized PageRank (PPR) diffusion based on the edge weight matrix of a GNNGraph and updates the graph with new edge weights derived from the PPR matrix.
 References paper: [The pagerank citation ranking: Bringing order to the web](http://ilpubs.stanford.edu:8090/422)
--- a/src/GNNGraphs/transform.jl
+++ b/src/GNNGraphs/transform.jl
@@ -1105,7 +1105,7 @@ function ppr_diffusion(g::GNNGraph{<:COO_T}; alpha = 0.85f0)
     N = g.num_nodes
 
     initial_A = sparse(t, s, w, N, N)
-    scaled_A = (alpha_f32 - 1) * initial_A
+    scaled_A = (Float32(alpha) - 1) * initial_A
 
     I_sparse = sparse(Diagonal(ones(Float32, N)))
     A_sparse = I_sparse + scaled_A
--- a/src/GNNGraphs/transform.jl
+++ b/src/GNNGraphs/transform.jl
@@ -1190,7 +1190,10 @@ The function performs the following steps:
 """
 function ppr_diffusion(g::GNNGraph{<:COO_T}; alpha = 0.85f0)
     s, t = edge_index(g)
-    w = ones(Float32, g.num_edges)
+    w = get_edge_weight(g)
+    if isnothing(w)
+        w = ones(Float32, g.num_edges)
+    end
 
     N = g.num_nodes
 
--- a/src/GNNGraphs/transform.jl
+++ b/src/GNNGraphs/transform.jl
@@ -1081,7 +1081,8 @@ ci2t(ci::AbstractVector{<:CartesianIndex}, dims) = ntuple(i -> map(x -> x[i], ci
 """
     ppr_diffusion(g::GNNGraph{<:COO_T}; alpha_f32::Float32=0.85f0) -> GNNGraph
 
-Calculates the Personalized PageRank (PPR) diffusion based on the edge weight matrix of a GNNGraph and updates the graph with new edge weights derived from the PPR matrix as introduced in the paper [The pagerank citation ranking: Bringing order to the web](http://ilpubs.stanford.edu:8090/422)
+Calculates the Personalized PageRank (PPR) diffusion based on the edge weight matrix of a GNNGraph and updates the graph with new edge weights derived from the PPR matrix.
+References paper: [The pagerank citation ranking: Bringing order to the web](http://ilpubs.stanford.edu:8090/422)
 
 
 The function performs the following steps:
--- a/src/GNNGraphs/transform.jl
+++ b/src/GNNGraphs/transform.jl
@@ -1079,7 +1079,7 @@ ci2t(ci::AbstractVector{<:CartesianIndex}, dims) = ntuple(i -> map(x -> x[i], ci
 @non_differentiable dense_zeros_like(x...)
 
 """
-    ppr_diffusion(g::GNNGraph; alpha_f32::Float32=0.85f0) -> GNNGraph
+    ppr_diffusion(g::GNNGraph{<:COO_T}; alpha_f32::Float32=0.85f0) -> GNNGraph
 
 Calculates the Personalized PageRank (PPR) diffusion based on the edge weight matrix of a GNNGraph and updates the graph with new edge weights derived from the PPR matrix.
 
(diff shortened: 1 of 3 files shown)


src/GNNGraphs/transform.jl
```diff
@@ -1078,3 +1078,53 @@ ci2t(ci::AbstractVector{<:CartesianIndex}, dims) = ntuple(i -> map(x -> x[i], ci
 @non_differentiable remove_self_loops(x...)  # TODO this is wrong, since g carries feature arrays, needs rrule
 @non_differentiable dense_zeros_like(x...)
 
+"""
+    ppr_diffusion(g::GNNGraph{<:COO_T}; alpha_f32::Float32=0.85f0) -> GNNGraph
+
+Calculates the Personalized PageRank (PPR) diffusion based on the edge weight matrix of a GNNGraph and updates the graph with new edge weights derived from the PPR matrix.
```
CarloLucibello, line 1084: a reference to a paper is needed

rbSparky: Done

rbSparky:
Tests fail for 
```
GATv2Conv: Test Failed at /home/runner/work/GraphNeuralNetworks.jl/GraphNeuralNetworks.jl/test/test_utils.jl:203
```

Will check

rbSparky:
Tests pass now, so the previous test fail was unrelated to this PR.

That should also be checked later.

src/GNNGraphs/transform.jl
```diff
@@ -1078,3 +1078,55 @@ ci2t(ci::AbstractVector{<:CartesianIndex}, dims) = ntuple(i -> map(x -> x[i], ci
...
+# Returns
+- A new `GNNGraph` instance with the same structure as `g` but with updated edge weights according to the PPR diffusion calculation.
+"""
+function ppr_diffusion(g::GNNGraph{<:COO_T}; alpha_f32::Float32=0.85f0)
+    s, t = edge_index(g)
+    w = get_edge_weight(g)  
+
+    N = g.num_nodes
+
+    A = zeros(Float32, N, N)
```
CarloLucibello, line 1107: 1. you should not assume the graph is on cpu; 2. you should not create dense matrices of size num_nodes x num_nodes, since num_nodes could be very large

src/GNNGraphs/transform.jl
```diff
@@ -1078,3 +1078,55 @@ ci2t(ci::AbstractVector{<:CartesianIndex}, dims) = ntuple(i -> map(x -> x[i], ci
...
+    
+    for (idx, src, dst) in zip(1:length(s), s, t)
+        A[dst, src] += w[idx]  
+    end
```
CarloLucibello, line 1111: single element operations are not gpu friendly

src/GNNGraphs/transform.jl
rbSparky: 1. what changes would need to be made for gpu support? what are the general guidelines?
2. will fix this

CarloLucibello: > you should not assume the graph is on cpu;
one option is `zeros_like(s, N, N)`

src/GNNGraphs/transform.jl
CarloLucibello: Actually you can just do 
```
A = adjacency_matrix(g::GNNGraph; dir=:out, weighted=true)
```
I'm not sure about the `dir` argument though, you should check which one (:in, :out, or :both) makes sense.

rbSparky: will do

rbSparky:
Would this be GPU compatible now that non mutable operations are removed? Tested training and Zygote no longer gives single element operation error!

CarloLucibello:
Let's have an implementation which works correctly with dense matrices on cpu and in later PRs we can add gpu support and add the sparse matrices performance optimization. So we are mostly fine with the current PR.

rbSparky:
> Let's have an implementation which works correctly with dense matrices on cpu and in later PRs we can add gpu support and add the sparse matrices performance optimization. So we are mostly fine with the current PR.

Should I change the code back to older commits where only dense matrices were used? Or would this be ok as it is?

test/GNNGraphs/transform.jl
```diff
@@ -519,4 +519,25 @@ end
...
     end
+end
+
+@testset "ppr_diffusion" begin
+    if GRAPH_T == :coo
+        s = [1, 1, 2, 3]
+        t = [2, 3, 4, 5]
+        eweights = [0.1, 0.2, 0.3, 0.4]
+
+        g = GNNGraph(s,t,eweights)
```
CarloLucibello, line 530:
```suggestion
        g = GNNGraph(s, t, eweights)
```

test/GNNGraphs/transform.jl
```diff
@@ -519,4 +519,25 @@ end
...
+        get_edge_weight(g)
+
+        g_new = ppr_diffusion(g)
+        w_new = get_edge_weight(g_new)
+
+        check_ew = Float32[0.012749999
+        0.025499998
+        0.038249996
+        0.050999995]
```
CarloLucibello, line 539: indentation

test/GNNGraphs/transform.jl
```diff
@@ -519,4 +519,25 @@ end
...
+end
+
+@testset "ppr_diffusion" begin
+    if GRAPH_T == :coo
+        s = [1, 1, 2, 3]
+        t = [2, 3, 4, 5]
+        eweights = [0.1, 0.2, 0.3, 0.4]
+
+        g = GNNGraph(s,t,eweights)
+        get_edge_weight(g)
```
CarloLucibello, line 531:
```suggestion
```

src/GNNGraphs/transform.jl
```diff
@@ -1078,3 +1078,46 @@ ci2t(ci::AbstractVector{<:CartesianIndex}, dims) = ntuple(i -> map(x -> x[i], ci
...
+# Arguments
+- `g::GNNGraph`: The input graph for which PPR diffusion is to be calculated. It should have edge weights available.
+- `alpha_f32::Float32`: The damping factor used in PPR calculation, controlling the teleport probability in the random walk. Defaults to `0.85f0`.
+
+# Returns
+- A new `GNNGraph` instance with the same structure as `g` but with updated edge weights according to the PPR diffusion calculation.
+"""
+function ppr_diffusion(g::GNNGraph{<:COO_T}; alpha_f32::Float32=0.85f0)
+    s, t = edge_index(g)
+    w = ones(Float32, g.num_edges)
```
CarloLucibello, line 1103: weights should be created from scratches only if the are not already present in`g`.

src/GNNGraphs/transform.jl
```diff
@@ -1078,3 +1078,46 @@ ci2t(ci::AbstractVector{<:CartesianIndex}, dims) = ntuple(i -> map(x -> x[i], ci
...
+4. Updates the original edge weights of the graph based on the PPR diffusion matrix, assigning new weights for each edge from the PPR matrix.
+
+# Arguments
+- `g::GNNGraph`: The input graph for which PPR diffusion is to be calculated. It should have edge weights available.
+- `alpha_f32::Float32`: The damping factor used in PPR calculation, controlling the teleport probability in the random walk. Defaults to `0.85f0`.
+
+# Returns
+- A new `GNNGraph` instance with the same structure as `g` but with updated edge weights according to the PPR diffusion calculation.
+"""
+function ppr_diffusion(g::GNNGraph{<:COO_T}; alpha_f32::Float32=0.85f0)
```
CarloLucibello, line 1101:
```suggestion
function ppr_diffusion(g::GNNGraph{<:COO_T}; alpha = 0.85f0)
```

src/GNNGraphs/transform.jl
```diff
@@ -1078,3 +1078,46 @@ ci2t(ci::AbstractVector{<:CartesianIndex}, dims) = ntuple(i -> map(x -> x[i], ci
...
+    s, t = edge_index(g)
+    w = ones(Float32, g.num_edges)
+
+    N = g.num_nodes
+
+    initial_A = sparse(t, s, w, N, N)
+    scaled_A = (alpha_f32 - 1) * initial_A
```
CarloLucibello, line 1108:
```suggestion
    scaled_A = (Float32(alpha) - 1) * initial_A
```

CarloLucibello:
bump. Also needs a rebase.

rbSparky:
> bump. Also needs a rebase.

in progress

test/GNNGraphs/transform.jl
```diff
@@ -595,4 +595,24 @@ end
...
+
+        g_new = ppr_diffusion(g)
+        w_new = get_edge_weight(g_new)
+
+        check_ew = Float32[0.012749999
+                           0.025499998
+                           0.038249996
+                           0.050999995]
+
+        isapprox(w_new, check_ew)
```
CarloLucibello, line 616:
```suggestion
        @test w_new ≈ check_ew
```

PR #429, opened 2024-03-19
Added `node_shuffle`

rbSparky:
Continues Issue #412 

[DGL function](https://docs.dgl.ai/en/0.8.x/generated/dgl.transforms.NodeShuffle.html)

Wasn't sure what to test in this one. Please let me know what other tests I should add since this is pretty straightforward.

### DIFF
--- a/src/GNNGraphs/transform.jl
+++ b/src/GNNGraphs/transform.jl
@@ -249,6 +249,28 @@ function remove_nodes(g::GNNGraph{<:COO_T}, nodes_to_remove::AbstractVector)
 end
 
 """
+    node_shuffle(g::GNNGraph{<:COO_T})
+
+Shuffle the node features in the given GNNGraph as shown in paper [On Node Features for Graph Neural Networks](https://arxiv.org/ftp/arxiv/papers/1911/1911.08795.pdf)
+
+# Arguments
+- `g::GNNGraph`: The input graph represented as a GNNGraph.
+
+# Returns
+- `g::GNNGraph`: The GNNGraph with shuffled node features.
+"""
+function node_shuffle(g::GNNGraph{<:COO_T})
+    features = g.ndata.x
+    num_nodes = size(features, 1) 
+
+    shuffled_indices = randperm(num_nodes)
+    shuffled_features = features[shuffled_indices, :]
+
+    g.ndata.x = shuffled_features
+    return g  
+end
+
+"""
     add_edges(g::GNNGraph, s::AbstractVector, t::AbstractVector; [edata])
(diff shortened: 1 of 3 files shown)


src/GNNGraphs/transform.jl
```diff
@@ -248,6 +248,28 @@ function remove_nodes(g::GNNGraph{<:COO_T}, nodes_to_remove::AbstractVector)
...
+
+Shuffle the node features in the given GNNGraph as shown in paper [On Node Features for Graph Neural Networks](https://arxiv.org/ftp/arxiv/papers/1911/1911.08795.pdf)
+
+# Arguments
+- `g::GNNGraph`: The input graph represented as a GNNGraph.
+
+# Returns
+- `g::GNNGraph`: The GNNGraph with shuffled node features.
+"""
+function node_shuffle(g::GNNGraph{<:COO_T})
```
CarloLucibello, line 262: this is mutating, so it should be called `node_shuffle!`

src/GNNGraphs/transform.jl
```diff
@@ -248,6 +248,28 @@ function remove_nodes(g::GNNGraph{<:COO_T}, nodes_to_remove::AbstractVector)
...
+    features = g.ndata.x
```
CarloLucibello, line 263: we can have no guarantess that there exists features named x

src/GNNGraphs/transform.jl
```diff
@@ -248,6 +248,28 @@ function remove_nodes(g::GNNGraph{<:COO_T}, nodes_to_remove::AbstractVector)
...
+    num_nodes = size(features, 1)
```
CarloLucibello, line 264: the node dimension is the second one

src/GNNGraphs/transform.jl
```diff
@@ -248,6 +248,28 @@ function remove_nodes(g::GNNGraph{<:COO_T}, nodes_to_remove::AbstractVector)
...
+
+# Returns
+- `g::GNNGraph`: The GNNGraph with shuffled node features.
+"""
+function node_shuffle(g::GNNGraph{<:COO_T})
+    features = g.ndata.x
+    num_nodes = size(features, 1) 
+
+    shuffled_indices = randperm(num_nodes)
+    shuffled_features = features[shuffled_indices, :]
```
CarloLucibello, line 267: this is shuffling feature dimensions, not nodes

CarloLucibello:
Besides the issues that I commented, I think that this function should not be part of the library. It is not a common operation, it has used in that paper for probing the relevance of node features. This kind of operations should be done on the user side.

PR #437, opened 2024-05-28
Add `GConvLSTM` temporal layer

aurorarossi:
This PR adds the `GConvLSTM` temporal convolutional layer of the following paper https://arxiv.org/pdf/1612.07659

### DIFF
--- a/src/layers/temporalconv.jl
+++ b/src/layers/temporalconv.jl
@@ -217,18 +217,22 @@ function GConvLSTMCell(ch::Pair{Int, Int}, k::Int, n::Int;
                         init = Flux.glorot_uniform,
                         init_state = Flux.zeros32)
     in, out = ch
+    # input gate
     conv_x_i = ChebConv(in => out, k; bias, init)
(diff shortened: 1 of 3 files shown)


aurorarossi:
In a future PR, I will update the documentation with a clear distinction between static graphs with temporal features convolution layers (like the one in this PR) and temporal graphs with temporal features layers.

src/layers/temporalconv.jl
```diff
@@ -187,6 +187,122 @@ function Base.show(io::IO, a3tgcn::A3TGCN)
...
+
+```jldoctest
+julia> g, x = rand_graph(5, 10), rand(Float32, 2, 5);
+
+julia> gclstm = GConvLSTM(2 => 5, 2, g.num_nodes);
+
+julia> y = gclstm(g, x);
+
+julia> size(y)
+(5, 5)
```
CarloLucibello, line 295: maybe the example could involve a temporal graph or a sequence of graphs

CarloLucibello (approved):

PR #445, opened 2024-07-12
Add `GCNConvFixedW` layer

aurorarossi:
This PR adds the version of `GCNConv` layer in which you have to pass the weights. 
It will be part of the temporal convolution EvolveGCN of the paper https://arxiv.org/abs/1902.10191.


CarloLucibello:
That's a lot of unfortunate code duplication. Can we just add another argument `weight = nothing` to the forward pass of the existing `GCNConv`?

aurorarossi:
See cleaner PR #447

PR #447, opened 2024-07-18
Add possibility to pass weights to `GCNConv`

aurorarossi:
Better solution of PR https://github.com/CarloLucibello/GraphNeuralNetworks.jl/pull/445

### DIFF
--- a/src/layers/conv.jl
+++ b/src/layers/conv.jl
@@ -40,7 +40,7 @@ and optionally an edge weight vector. Returns a node feature matrix of size
 
 The `norm_fn` parameter allows for custom normalization of the graph convolution operation by passing a function as argument. 
 By default, it computes ``\frac{1}{\sqrt{d}}`` i.e the inverse square root of the degree (`d`) of each node in the graph. 
-If `conv_weight` is an `AbstractMatrix` of size `[out, in]`, then the convolution is performed using that weight matrix.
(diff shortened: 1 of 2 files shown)


src/layers/conv.jl
```diff
@@ -32,14 +32,15 @@ and optionally an edge weight vector.
...
+    (::GCNConv)(g::GNNGraph, x::AbstractMatrix, edge_weight = nothing, norm_fn::Function = d -> 1 ./ sqrt.(d), conv_weight::Union{Nothing,AbstractMatrix} = nothing) -> AbstractMatrix
 
-Takes as input a graph `g`,ca node feature matrix `x` of size `[in, num_nodes]`,
+Takes as input a graph `g`, a node feature matrix `x` of size `[in, num_nodes]`,
 and optionally an edge weight vector. Returns a node feature matrix of size 
 `[out, num_nodes]`.
 
 The `norm_fn` parameter allows for custom normalization of the graph convolution operation by passing a function as argument. 
 By default, it computes ``\frac{1}{\sqrt{d}}`` i.e the inverse square root of the degree (`d`) of each node in the graph. 
+If `conv_weight` is an `AbstractMatrix` of size `[out, in]`, then the convolution is performed using that weight matrix.
```
CarloLucibello, line 43:
```suggestion
If `conv_weight` is an `AbstractMatrix` of size `[out, in]`, then the convolution is performed using that weight matrix instead of the weights stored in the model.
```

PR #470, opened 2024-07-29
[GNNLux] TGCN temporal layer

aurorarossi:
This PR implements the `TGCN` layer in Lux.jl. I implemented it directly in Lux without implementing the `GNNlib` function because the recurrence layers in Lux are different from those in Flux.

### DIFF
--- a/GNNLux/src/layers/temporalconv.jl
+++ b/GNNLux/src/layers/temporalconv.jl
@@ -2,21 +2,25 @@
     cell <: Union{<:Lux.AbstractRecurrentCell, <:GNNContainerLayer}
 end
 
-function initialstates(rng::AbstractRNG, r::StatefulRecurrentCell)
-    return (cell=initialstates(rng, r.cell), carry=nothing)
+function LuxCore.initialstates(rng::AbstractRNG, r::GNNLux.StatefulRecurrentCell)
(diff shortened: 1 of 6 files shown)


GNNLux/src/layers/temporalconv.jl
```diff
@@ -0,0 +1,59 @@
...
+    (out, carry), st = applyrecurrentcell(r.cell, g, x, ps, st.cell, st.carry)
+    return out, (; cell=st, carry)
+end
+
+function (r::StatefulRecurrentCell)(g, x::AbstractVector, ps, st::NamedTuple)
+    (out, carry), st = applyrecurrentcell(r.cell, g, first(x), ps, st.cell, st.carry)
+    for xᵢ in x[(begin + 1):end]
+        (out, carry), st = applyrecurrentcell(r.cell, g, xᵢ, ps, st, carry)
+    end
+    return out, (; cell=st, carry)
```
CarloLucibello, line 19:
```suggestion
    st, carry = st.cell, st.carry
    for xᵢ in x
        (out, carry), st = applyrecurrentcell(r.cell, g, xᵢ, ps, st, carry)
    end
    return out, (; cell=st, carry)
```

PR #474, opened 2024-08-01
[GNNGraphs] implement `remove_edges(g, p)`

rbSparky:
This is the second step in successfully implementing Issue https://github.com/CarloLucibello/GraphNeuralNetworks.jl/issues/412 for which we will need to implement [DropEdge](https://arxiv.org/abs/1907.10903).

### DIFF
--- a/GNNGraphs/src/transform.jl
+++ b/GNNGraphs/src/transform.jl
@@ -151,7 +151,7 @@ end
 
 """
     remove_edges(g::GNNGraph, edges_to_remove::AbstractVector{<:Integer})
-    remove_edges(g::GNNGraph, p::Float64=0.5)
+    remove_edges(g::GNNGraph, p=0.5)
 
 Remove specified edges from a GNNGraph, either by specifying edge indices or by randomly removing edges with a given probability.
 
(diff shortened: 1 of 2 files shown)


GNNGraphs/src/transform.jl
```diff
@@ -151,12 +151,14 @@ end
 
 """
     remove_edges(g::GNNGraph, edges_to_remove::AbstractVector{<:Integer})
+    remove_edges(g::GNNGraph, p::Float64=0.5)
```
CarloLucibello, line 154:
```suggestion
    remove_edges(g::GNNGraph, p=0.5)
```
no need to type restrict (especially since we don't to it in the signature).

GNNGraphs/src/transform.jl
```diff
@@ -200,6 +210,20 @@ function remove_edges(g::GNNGraph{<:COO_T}, edges_to_remove::AbstractVector{<:In
...
+    num_edges = g.num_edges
+    edges_to_remove = filter(_ -> rand() < p, 1:num_edges)        
+    g = remove_edges(g, edges_to_remove)
+    s, t = edge_index(g)
+    w = get_edge_weight(g)
+    edata = g.edata
+    return GNNGraph((s, t, w),
+             g.num_nodes, length(s), g.num_graphs,
+             g.graph_indicator,
+             g.ndata, edata, g.gdata)
```
CarloLucibello, line 224: there is any reason for not doing simply
```suggestion
    return remove_edges(g, edges_to_remove)
```
?

rbSparky: Yes, my bad

Issue #461, opened 2024-07-26
plan for GNNLux.jl


CarloLucibello:
- [x] create the package #460 
- [ ] make it feature parity with GraphNeuralNetworks.jl (#469, #463, #475, #471)
    - [x] GNNChain
    - [x] GCNConv
    - [x] GraphConv 
    - [x] ChebConv
    - [x] AGNNConv
    - [x] CGConv
    - [x] EGNNConv
    - [x] EdgeConv
    - [x] GATConv
    - [x] GATv2Conv
    - [x] GatedGraphConv
    - [x] GINConv
    - [x] GMMConv
    - [x] MEGNetConv
    - [x] NNConv
    - [x] ResGatedGraphConv
    - [x] SAGEConv
    - [x] SGConv
    - [ ] TAGConv
    - [ ] TransformerConv
- [x] temporal convs (#470 )
- [ ] create documentation 
- [ ] Consider alternative names?
- [ ] register the package

cc @avik-pal

asinghvi17:
@CarloLucibello sorry to ping here, but what's the status on GNNLux.jl these days?  Is it close to release / is there anything one could help with to move that over the finish line?

CarloLucibello:
We are almost there, very few layers left. We need to add a docstring to each layer, adapting the ones from GraphNeuralNetworks.jl. If you want to help with that it would be great!

asinghvi17:
Will give it a look this week!!

PR #475, opened 2024-08-01
[GNNLux] Added SGConv 

rbSparky:
Adding conv layer according to [plan for GNNLux](https://github.com/CarloLucibello/GraphNeuralNetworks.jl/issues/461)


GNNlib/src/layers/conv.jl
```diff
@@ -723,3 +723,57 @@ function d_conv(l, g::GNNGraph, x::AbstractMatrix)
     end
     return h .+ l.bias
 end
+
+####################### SGConv ######################################
+
+function sg_conv(l, g::AbstractGNNGraph, x, edge_weight::EW, conv_weight::CW) where
```
CarloLucibello, line 729: the `sg_conv` function already exists, all this should be removed

GNNLux/test/layers/conv_tests.jl
```diff
@@ -77,5 +77,10 @@
 
         #TODO test edge
     end
+
+    @testset "SGConv" begin
+        l = SGConv(in_dims => out_dims, relu)
```
CarloLucibello, line 82: the second argument to the constructor is not an activation function. 
I suggest running GNNLux tests locally (they are quite fast) and make sure they pass.

GNNLux/src/layers/conv.jl
```diff
@@ -515,4 +515,60 @@ function Base.show(io::IO, l::GATv2Conv)
...
+end
+
+function SGConv(ch::Pair{Int, Int}, k = 1;
+                init_weight = glorot_uniform,
+                init_bias = zeros32,
+                use_bias::Bool = true,
+                add_self_loops::Bool = true,
+                use_edge_weight::Bool = false)
+    in_dims, out_dims = ch
+    return SGConv(in_dims, out_dims, use_bias, add_self_loops, use_edge_weight, init_weight, init_bias, k)
```
CarloLucibello, line 538: the arguments passed to the constructor are not in the correct order. Please be more careful before committing and run tests locally. I cannot go over every tiny detail.

rbSparky: will check

GNNLux/test/layers/conv_tests.jl
rbSparky: will do

rbSparky:
Fixing tests

GNNLux/src/layers/conv.jl
```diff
@@ -515,4 +515,60 @@ function Base.show(io::IO, l::GATv2Conv)
...
+    l.use_bias || print(io, ", use_bias=false")
+    l.add_self_loops || print(io, ", add_self_loops=false")
+    !l.use_edge_weight || print(io, ", use_edge_weight=true")
+    print(io, ")")
+end
+
+(l::SGConv)(g, x, ps, st; conv_weight=nothing, edge_weight=nothing) = 
+    l(g, x, edge_weight, ps, st; conv_weight)
+
+function (l::SGConv)(g, x, edge_weight, ps, st;
```
CarloLucibello, line 567: this actually should be
```julia
(l::SGConv)(g, x, ps, st) = l(g, x, nothing, ps, st)

function (l::SGConv)(g, x, edge_weight, ps, st)
```
Fixed it in #471

PR #491, opened 2024-09-14
[GNNLux] Adding NNConv Layer 

rbSparky:

### DIFF
--- a/GNNLux/test/shared_testsetup.jl
+++ b/GNNLux/test/shared_testsetup.jl
@@ -14,7 +14,7 @@ export test_lux_layer
 
 function test_lux_layer(rng::AbstractRNG, l, g::GNNGraph, x; 
             outputsize=nothing, sizey=nothing, container=false,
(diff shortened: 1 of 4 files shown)


rbSparky:
Tests pass

GNNLux/test/shared_testsetup.jl
```diff
@@ -14,7 +14,7 @@ export test_lux_layer
 
 function test_lux_layer(rng::AbstractRNG, l, g::GNNGraph, x; 
             outputsize=nothing, sizey=nothing, container=false,
-            atol=1.0f-2, rtol=1.0f-2)
+            atol=1.0f-2, rtol=1.0f-2, edge_weight=nothing)
```
CarloLucibello, line 17: Let's call it `e` instead of `edge_weight` since we reserve the term edge weight to scalars associated to each edge while these are edge features instead

CarloLucibello:
The show method is missing

rbSparky:
Show method added. Changed `edge_weight` to `e`

PR #501, opened 2024-09-30
[WIP] [GNNLux] Adding TransformerConv Layer

rbSparky:


rbSparky:
tests work with skip connection false

but fail with skip connection true

dimension error:
```
Error in testset "TransformerConv" on worker 2157:
Error During Test at /home/runner/work/GraphNeuralNetworks.jl/GraphNeuralNetworks.jl/GNNLux/test/layers/conv_tests.jl:8
Got exception outside of a @test
DimensionMismatch: A has shape (16, 16) but B has shape (6, 10)
```

rbSparky:
@CarloLucibello let me know if you have ideas to fix with skip connection
I will look further into it too

PR #521, opened 2024-11-12
Add `GNNLux` training example in docs

aurorarossi:


CarloLucibello:
We could add DataLoader usage to the example

aurorarossi:
> We could add DataLoader usage to the example

Yes, but at the moment it is not straight forward.

PR #543, opened 2024-12-05
Add version selector

aurorarossi:
Fix https://github.com/JuliaGraphs/GraphNeuralNetworks.jl/issues/536

### DIFF
--- a/.github/workflows/multidocs.yml
+++ b/.github/workflows/multidocs.yml
@@ -86,7 +86,7 @@ jobs:
       - name: Build multidocs
         if: github.event_name == 'pull_request'
         run: |
-          julia --project=docs/ docs/make-multi.jl 
+          julia --project=docs/ docs/make-multi.jl PR
 
       - name: Build and deploy multidocs
         if: github.event_name != 'pull_request'


docs/make-multi.jl
```diff
@@ -39,11 +39,10 @@ outpath = joinpath(@__DIR__, "build")
...
     docs;
-    search_engine = MultiDocumenter.SearchConfig(),
-    # search_engine = MultiDocumenter.SearchConfig(
-    #     index_versions = ["stable"],
-    #     engine = MultiDocumenter.FlexSearch
-    # ),
+    search_engine = MultiDocumenter.SearchConfig(
+        index_versions = ["stable"],
+        engine = MultiDocumenter.FlexSearch
+    ),
```
CarloLucibello, line 45: it would be good to remove the search entirely

CarloLucibello (approved):

docs/make-multi.jl
aurorarossi: before we removed it, but the bar was still there. Perhaps the bar would be useful if versioning worked.

PR #551, opened 2024-12-07
Stable documentation

aurorarossi:


CarloLucibello:
What is this doing?

aurorarossi:
I have adjusted the workflow so that the documentation refers to the master branch. As you can see, previously the workflow was using the dev version of the package.

CarloLucibello:
with this we loose developer docs though

CarloLucibello:
We can close this, with #552 there has been some progress

Issue #571, opened 2024-12-26
add reference to PyGDatasets.jl


CarloLucibello:
https://github.com/CarloLucibello/PyGDatasets.jl

PR #577, opened 2025-01-11
remove explicit dependence on GPUArraysCore

axsk:


CarloLucibello:
Let's remove the dependence on GPUArraysCore entirely, it is not needed

axsk:
Removed

CarloLucibello:
we do need the one in the test/Project.toml though, because of https://github.com/JuliaGPU/CUDA.jl/issues/2564

CarloLucibello:
done in #571

PR #601, opened 2025-06-16
Provide CUDA support for the binarize() operation on sparse matrices

dferre97:
Made _binarize()_ work on CUDA with sparse adjmat of type CUSPACE.CuSparseMatrixCSC.
Also, added a specialization of _binarize()_ that takes an eltype as an additional arg and directly creates a binarized adjmat of that, avoiding potentially costly subsequent conversions.
However, note that for now this implementation won't be called during the fwd pass, as the CUDA propagate specialization falls back on the default gather/scatter approach, which doesn't use the adjecency matrix.
All tests on CPU and GPU correctly passed, except for 1 broken in "ChebConv GPU", but it is also broken on master.

### DIFF
--- a/GNNGraphs/ext/GNNGraphsCUDAExt.jl
+++ b/GNNGraphs/ext/GNNGraphsCUDAExt.jl
@@ -4,6 +4,7 @@ using CUDA
 using Random, Statistics, LinearAlgebra
 using GNNGraphs
 using GNNGraphs: COO_T, ADJMAT_T, SPARSE_T 
+using SparseArrays
 
 const CUMAT_T = Union{CUDA.AnyCuMatrix, CUDA.CUSPARSE.CuSparseMatrix}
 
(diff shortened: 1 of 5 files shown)


GNNGraphs/ext/GNNGraphsCUDAExt.jl
```diff
@@ -20,6 +21,18 @@ GNNGraphs.dense_zeros_like(a::CUMAT_T, T::Type, sz = size(a)) = CUDA.zeros(T, sz
 
 GNNGraphs.iscuarray(x::AnyCuArray) = true
 
+function GNNGraphs.binarize(Mat::CUSPARSE.CuSparseMatrixCSC)
+    @debug "Binarizing CuSparseMatrixCSC"
```
CarloLucibello, line 25:
```suggestion
```

GNNGraphs/ext/GNNGraphsCUDAExt.jl
```diff
@@ -20,6 +21,18 @@ GNNGraphs.dense_zeros_like(a::CUMAT_T, T::Type, sz = size(a)) = CUDA.zeros(T, sz
...
+    bin_vals = fill!(similar(nonzeros(Mat), Bool), true)
+    return CUSPARSE.CuSparseMatrixCSC(Mat.colPtr, rowvals(Mat), bin_vals, size(Mat))
+end
+function GNNGraphs.binarize(Mat::CUSPARSE.CuSparseMatrixCSC, T::DataType)
+    @debug "Binarizing CuSparseMatrixCSC with type $(T)"
```
CarloLucibello, line 30:
```suggestion
```

GNNGraphs/src/utils.jl
```diff
@@ -296,6 +296,8 @@ function _rand_edges(rng, (n1, n2), m)
 end
 
 binarize(x) = map(>(0), x)
+# here just to allow CUDA extension to overload this function with correct type casting
+binarize(x, T::DataType) = binarize(x)
```
CarloLucibello, line 300:
```suggestion
binarize(x, T::DataType) = map(y -> y > 0 ? one(T) : zero(T), x)
```
or something similar

CarloLucibello:
The segmentation fault in GNNGGraphs CUDA CI should be investigated

GNNGraphs/ext/GNNGraphsCUDAExt.jl
```diff
@@ -20,6 +21,16 @@ GNNGraphs.dense_zeros_like(a::CUMAT_T, T::Type, sz = size(a)) = CUDA.zeros(T, sz
 
 GNNGraphs.iscuarray(x::AnyCuArray) = true
 
+function GNNGraphs.binarize(Mat::CUSPARSE.CuSparseMatrixCSC)
+    bin_vals = fill!(similar(nonzeros(Mat), Bool), true)
+    return CUSPARSE.CuSparseMatrixCSC(Mat.colPtr, rowvals(Mat), bin_vals, size(Mat))
+end
```
CarloLucibello, line 27:
```suggestion
```
This could be removed in favor of a single method with signature `binarize(Mat::CUSPARSE.CuSparseMatrixCSC, T::DataType = Bool). Same for the main method in GNNGraphs I guess.

PR #605, opened 2025-07-03
CUDA copy_xj propagate sparse support and benchmarks

dferre97:
Added fast copy_xj propagate CUDA support for sparse graphs using SpMM.
Added benchmarks to compare with gather/scatter approach, speedup of 100-200x on average.

### DIFF
--- a/GNNlib/ext/GNNlibCUDAExt.jl
+++ b/GNNlib/ext/GNNlibCUDAExt.jl
@@ -7,14 +7,6 @@ using GNNGraphs: GNNGraph, COO_T, SPARSE_T
 
 ###### PROPAGATE SPECIALIZATIONS ####################
 
-## COPY_XJ 
-
-## avoid the fast path on gpu until we have better cuda support
-function GNNlib.propagate(::typeof(copy_xj), g::GNNGraph{<:Union{COO_T, SPARSE_T}}, ::typeof(+),
-        xi, xj::AnyCuMatrix, e)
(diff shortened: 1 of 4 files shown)


GNNlib/ext/GNNlibCUDAExt.jl
```diff
@@ -7,14 +7,6 @@ using GNNGraphs: GNNGraph, COO_T, SPARSE_T
...
 ###### PROPAGATE SPECIALIZATIONS ####################
 
-## COPY_XJ 
-
-## avoid the fast path on gpu until we have better cuda support
-function GNNlib.propagate(::typeof(copy_xj), g::GNNGraph{<:Union{COO_T, SPARSE_T}}, ::typeof(+),
-        xi, xj::AnyCuMatrix, e)
-    propagate((xi, xj, e) -> copy_xj(xi, xj, e), g, +, xi, xj, e)
-end
-
```
CarloLucibello, line 17: We should keep this, but only for the case coo_t

GNNlib/src/msgpass.jl
```diff
@@ -213,7 +213,7 @@ end
 ## COPY_XJ 
 
 function propagate(::typeof(copy_xj), g::GNNGraph, ::typeof(+), xi, xj::AbstractMatrix, e)
-    A = adjacency_matrix(g, weighted = false)
+    A = adjacency_matrix(g, eltype(xj); weighted = false)
```
CarloLucibello, line 216: Is the cast to the xj type necessary?

PR #610, opened 2025-07-16
Add propagate w_mul_xj CUDA sparse support using matrix mul

dferre97:
Added fast w_mul_xj propagate CUDA support for sparse graphs using SpMM.
Added benchmarks to compare with gather/scatter approach, speedup from 40x to 300x on my machine, huge memory allocation benefits (up to 1000x less or more depending on size of graph and sparsity level).
CUDA tests on GraphNeuralNetworks passed.


CarloLucibello:
Build kite Cuda issues doesn't seem related to this pr, but we have to fix it at some point.

PR #634, opened 2025-12-24
feat: Reduce nodes over heterogeneous graphs

### LINKED ISSUE #622: Pooling over heterographs
Is there a way to make the `GlobalPool` layer work for `GNNHeteroGraph`s? I only need to pool over one kind of nodes.

My current solution:
```julia
x = NNlib.scatter(model.pool.aggr, x[:kind], graph_indicator(g, :kind))
```

Is there a way to pool over all types of nodes or some types of nodes?

lenianiva:
Resolves #622

### DIFF
--- a/GNNlib/test/utils.jl
+++ b/GNNlib/test/utils.jl
@@ -19,6 +19,17 @@
         @test r2 == r
     end
 
+    @testset "reduce_nodes" begin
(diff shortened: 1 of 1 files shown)


CarloLucibello:
looks good! can you add tests?

CarloLucibello:
> Do you know why the build is failing?

don't worry about that one, it was already failing

PR #642, opened 2026-03-07
Test: Replaced raw Mooncake with Flux implementation and checked Mooncake compatibility for more layers

Parvm1102:
I had reported the friendly_tangents= true issue to Mooncake.jl and they have resolved it in Mooncake v0.5.17.
I have changed the implementation with the preferred Flux.withgradient().

Also I have also checked few more layers in basic.jl, pool.jl and temporal.jl.
Observations:

- GlobalPool:	PASS
- GlobalAttentionPool: PASS
- TGCNCell / TGCN: PASS
- EvolveGCNO: PASS
- GNNChain: Fail	(MooncakeRuleCompilationError, complex chain with anonymous fns, Dropout, BatchNorm)
- GConvLSTMCell / GConvLSTM: Fail (Uses ChebConv internally)
- GConvGRUCell / GConvGRU: Fail (Uses ChebConv internally)
- DCGRUCell / DCGRU: Fail (Uses DConv internally)
- EvolveGCNOCell: Fail ("with state" call fails during Mooncake rule compilation)

### DIFF
--- a/GraphNeuralNetworks/test/test_module.jl
+++ b/GraphNeuralNetworks/test/test_module.jl
@@ -155,9 +155,7 @@ function test_gradients(
 
         if test_mooncake
             # Mooncake gradient with respect to f via Flux integration, compared against Zygote.
-            result = Flux.withgradient(f -> loss(f, graph, xs...), Flux.AutoMooncake(), f)
-            y_mc = result.val  # Extract value from NamedTuple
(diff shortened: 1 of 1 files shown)


CarloLucibello:
> GNNChain: Fail (MooncakeRuleCompilationError, complex chain with anonymous fns, Dropout, BatchNorm)

It would be good to have a MWE for this and report it upstream

PR #659, opened 2026-03-28
Fix/remove self loops adjmat mutation

Parvm1102:
closes #658 

Now the graph will be copied so that it does not mutate the original graph in place.

I have also added tests for remove_self_loops for dense graph as well


Parvm1102:
@CarloLucibello , the GNNLux checks are failing, they were also failing in the last few open prs. 
I think it is related to some recent upstream changes thats triggering Automooncake() tests.

PR #664, opened 2026-03-30
Add ZINC graph regression example

Uneeb808:


CarloLucibello:
please don't flood the repo with examples that add nothing with respect to those we already have

PR #665, opened 2026-04-04
Skip Mooncake tests on GNNLux and Sparse graphs

Parvm1102:
Recently, the GNNLux tests were failing in the CI because the newer version `LuxTestUtils v2.3.0` was also checking for AutoMooncake() AD. Most of the layers in GNNLux are not compatible with mooncake. I skipped mooncake for those layers.

### DIFF
--- a/GNNLux/test/layers/temporalconv.jl
+++ b/GNNLux/test/layers/temporalconv.jl
@@ -1,6 +1,6 @@
 @testitem "layers/temporalconv" setup=[TestModuleLux] begin
     using .TestModuleLux
-    using LuxTestUtils: test_gradients, AutoTracker, AutoForwardDiff, AutoEnzyme
+    using LuxTestUtils: test_gradients, AutoTracker, AutoForwardDiff, AutoEnzyme, AutoMooncake
 
     rng = StableRNG(1234)
     g = rand_graph(rng, 10, 40)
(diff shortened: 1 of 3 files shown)


Parvm1102:
The layers incompatible with Mooncake in GNNLux:

- Basic: GNNChain
- Conv: GCNConv, ChebConv, GraphConv, AGNNconv, EdgeConv, DConv, GATConv, GATv2Conv, SGConv, GatedGraphConv, GINConv, NNConv, ResGatedGraphConv and SAGEConv.
- Pool: GlobalAttentionPool
- temporalconv: All layers failed

Parvm1102:
@CarloLucibello , Now the GNN / julia 1 test is failing even though nothing is changed in the codebase, the error is as same as the one on GNNLux failure. 
This is unexpected because the same test was passing on the last pr. I am currently investigating the issue.

Parvm1102:
@CarloLucibello , I tried to find the cause of GNN / julia 1 test failing. 
This issue is most likely related to Mooncake v0.5.25 which was released 3 days ago. The PRs were passing tests before that.

It fails with: 
` MethodError: no method matching zero(::Mooncake.FriendlyTangentCache{Mooncake.AsRaw, Nothing})`

The failure is not layer specific. It is related to how the friendly-tangent is changed in v0.5.25. might be related to. 
should I try reporting to Mooncake.jl? What should be done with the failing tests as of now?

CarloLucibello:
Let's try to fix the Mooncake tests on GNN Flux. If it cannot be done quickly, let's just skip the Mooncake tests for the time being so that we have CI giving green lights again.

Parvm1102:
I have skipped the Mooncake testing just for sparse arrays in Flux. This way we don't need to block all tests.
The main reason of failure was friendly_tangents crashed on SparseMatrix with non-float type due to zero, should I report this to Mooncake.jl?

@CarloLucibello please review this pr.

PR #672, opened 2026-05-05
Avoid scalar indexing in remove_edges

### LINKED ISSUE #668: Scalar indexing triggered by remove_edges
The function `remove_edges` triggers a scalar indexing exception when the graph resides on the GPU. Other similar functions, like `add_edges`, don't have this problem

A MWE using CUDA:
```julia
using CUDA, GraphNeuralNetworks, SparseArrays, Flux
G = GNNGraph(A, ndata=randn(20)) |> Flux.gpu

julia> remove_edges(G, cu([3, 4, 5]))
ERROR: Scalar indexing is disallowed.
Invocation of getindex resulted in scalar indexing of a GPU array.
This is typically caused by calling an iterating implementation of a method.
Such implementations *do not* execute on the GPU, but very slowly on the CPU,
and therefore should be avoided.
```

GraphNeuralNetworks version is v1.1.0. 

~Could it be that this method is specialized on `GNNGraph{<:COO_t}`? From a quick look at the function's source, it doesn't look like it does anything GPU-illegal~

The culprit seems to be:
```julia
mask_to_keep = trues(length(s))
mask_to_keep[edges_to_remove] .= false # This line
```
This broadcast is not an issue when `mask_to_keep` is a GPU vector as well

alonsoC1s:
Fixes #668 by creating a boolean mask on GPU when the inputs are, without modifying the effects in normal CPU mode

### DIFF
--- a/GNNGraphs/src/GNNGraphs.jl
+++ b/GNNGraphs/src/GNNGraphs.jl
@@ -2,7 +2,7 @@ module GNNGraphs
 
 using SparseArrays
 import Graphs
-using Graphs: AbstractGraph, outneighbors, inneighbors, adjacency_matrix, degree, 
+using Graphs: AbstractGraph, outneighbors, inneighbors, adjacency_matrix, degree,
               has_self_loops, is_directed, induced_subgraph, has_edge
 import NearestNeighbors
 import NNlib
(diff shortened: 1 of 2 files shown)


alonsoC1s:
Failures seem unrelated

CarloLucibello:
Can you add a gpu test in GNNGraphs/test/transform.jl?

alonsoC1s:
Of course. Is it enough to add them to that file for the GPU code to be tested, or should I modify the testing configuration/flags/functions?

PR #681, opened 2026-07-07
fix: made the gradient testing more generic

### LINKED ISSUE #676: Make the gradient testing function more generic
It should take generic AD types. Right know it is Zygote based:
https://github.com/JuliaGraphs/GraphNeuralNetworks.jl/blob/f09fcc4af170c1d87dc1d51648c3a015357fea93/GraphNeuralNetworks/test/test_module.jl#L81
https://github.com/JuliaGraphs/GraphNeuralNetworks.jl/blob/f09fcc4af170c1d87dc1d51648c3a015357fea93/GNNlib/test/test_module.jl#L75

We should adapt the one from Flux
https://github.com/FluxML/Flux.jl/blob/57e29baf48edf50c0dd4fc9f027a8900ce3cef66/test/test_module.jl#L105

Parvm1102:
closes #676 

The new code passes all tests.
It has been made more generic by using a helper function for running the required backend.

### DIFF
--- a/GNNlib/test/msgpass.jl
+++ b/GNNlib/test/msgpass.jl
@@ -132,14 +132,14 @@ end
     @testset "copy_xj +" begin
         for g in TEST_GRAPHS
             f(g, x) = propagate(copy_xj, g, +, xj = x)
-            test_gradients(f, g, g.x; test_grad_f=false, test_mooncake=TEST_MOONCAKE)
+            test_gradients(f, g, g.x; test_grad_f=false, ad_backends=AD_BACKENDS)
         end
     end
 
     @testset "copy_xj mean" begin
         for g in TEST_GRAPHS
             f(g, x) = propagate(copy_xj, g, mean, xj = x)
-            test_gradients(f, g, g.x; test_grad_f=false, test_mooncake=TEST_MOONCAKE)
+            test_gradients(f, g, g.x; test_grad_f=false, ad_backends=AD_BACKENDS)
         end
     end
 
@@ -147,7 +147,7 @@ end
         for g in TEST_GRAPHS
             e = rand(Float32, size(g.x, 1), g.num_edges)
             f(g, x, e) = propagate(e_mul_xj, g, +; xj = x, e)
-            test_gradients(f, g, g.x, e; test_grad_f=false, test_mooncake=TEST_MOONCAKE)
+            test_gradients(f, g, g.x, e; test_grad_f=false, ad_backends=AD_BACKENDS)
         end
     end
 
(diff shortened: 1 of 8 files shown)


Parvm1102:
@CarloLucibello please review this

GNNlib/test/test_module.jl
```diff
@@ -96,6 +103,12 @@ function test_gradients(
                CPU AD vs GPU AD, or test Mooncake AD.")
     end
 
+    # AD backends to compare against the Zygote reference gradient.
+    ad_backends = []
```
CarloLucibello: this should be an argument to the function, with default AutoZygote. 

see how it is done here
https://github.com/FluxML/NNlib.jl/blob/2e687800a16391d58fefba2b929b19f5a2509c72/test/test_module.jl#L129

Parvm1102:
@CarloLucibello I have made some changes, does that look alright? 
Note: we would still need to pass ad_backends because there are some layers which fail mooncake. So we need some kind of flag or an array for tests

GraphNeuralNetworks/test/test_module.jl
```diff
@@ -87,13 +97,18 @@ function test_gradients(
             test_grad_f = true,
             test_grad_x = true,
             compare_finite_diff = true,
```
CarloLucibello:
```suggestion

```

CarloLucibello: this option should go. As in NNlib
https://github.com/FluxML/NNlib.jl/blob/2e687800a16391d58fefba2b929b19f5a2509c72/test/test_module.jl#L128
we should have a reference AD, which is by default FiniteDifferences for cpu tests.

GraphNeuralNetworks/test/test_module.jl
```diff
@@ -87,13 +97,18 @@ function test_gradients(
...
-            test_mooncake = false,
+            ad_backends = [],
```
CarloLucibello:
```suggestion
            ad_backends = [AutoZygote(), AutoMooncake()]
```
Then inside the function we remove AutoMooncake from the list if a global option NNLIB_TEST_MOONCAKE is false.

GNNlib/test/msgpass.jl
```diff
@@ -132,22 +132,22 @@ end
     @testset "copy_xj +" begin
         for g in TEST_GRAPHS
             f(g, x) = propagate(copy_xj, g, +, xj = x)
-            test_gradients(f, g, g.x; test_grad_f=false, test_mooncake=TEST_MOONCAKE)
+            test_gradients(f, g, g.x; test_grad_f=false, ad_backends=AD_BACKENDS)
```
CarloLucibello:
```suggestion
            test_gradients(f, g, g.x; test_grad_f=false)
```
No need to pass AD_BACKENDS anywhere (and we actually don't even need to define it). Just set as default `[AutoZygote(), AutoMooncake()]` in `test_gradients`.

Parvm1102: but then how do we skip tests for layers which are failing under mooncake?

CarloLucibello: only for those we explicitly pass `ad_backends=AutoZygote()`.

Parvm1102:
@CarloLucibello waiting for your review

PR #703, opened 2026-08-01
Fix Enzyme differentiation of GCNConv/SGConv/TAGConv on adjacency-matrix graphs via a keyword-free COO conversion helper

Parvm1102:
Fixes `Enzyme.gradient` crashing (`EnzymeInternalError`, EnzymeAD/Enzyme.jl#3409) when
differentiating `GCNConv`, `SGConv` or `TAGConv` on `:dense`/`:sparse` adjacency graphs.

The crash itself was fixed upstream in Enzyme 0.13.197 (EnzymeAD/Enzyme.jl#3422, EnzymeAD/Enzyme.jl#3424). The one
thing Enzyme still cannot compile is the union-typed keyword handling of `GNNGraph(g; graph_type)`
and `to_coo`, so this PR:

- adds an internal keyword-free helper `GNNGraphs._to_coo_graph(g)` that converts an
  adjacency-matrix graph to COO through the positional inner constructor;
- uses it in the adjacency-matrix fallbacks of `gcn_conv`, `sgc_conv`, `sg_conv`, `tag_conv`.

With this, `Enzyme.gradient` through the conversion matches finite differences on Julia 1.10 and
1.12 (no version gate needed with Enzyme ≥ 0.13.197), edge weights stay differentiable, and
Zygote/Mooncake behavior is unchanged. Tested end-to-end: `GCNConv`/`SGConv` gradients on
`:dense`/`:sparse` match Zygote.

Versions: GNNGraphs → 1.5.2-DEV, GNNlib → 1.4.1-DEV (requires GNNGraphs ≥ 1.5.2).


wsmoses:
can you open a separate issue for the 1.10 failure?

GNNGraphs/ext/GNNGraphsEnzymeCoreExt.jl
```diff
@@ -0,0 +1,10 @@
+module GNNGraphsEnzymeCoreExt
+
+using GNNGraphs
+using EnzymeCore: EnzymeRules
+
+# Storage conversion is structural and non-differentiable, matching its ChainRules
+# treatment; the rule also keeps Enzyme out of Union-heavy code it cannot compile.
+EnzymeRules.inactive(::typeof(GNNGraphs._to_coo_graph), ::GNNGraph) = nothing
```
wsmoses: just because it is structural doesn't mean it is inactive (e.g. the correct answer here as the rule may be to call _to_coo_graph on the primal and the shadow)

GNNGraphs/src/convert.jl
```diff
@@ -116,6 +116,19 @@ function to_coo(adj_list::ADJLIST_T; dir = :out, num_nodes = nothing, weighted =
     (s, t, nothing), num_nodes, num_edges
 end
 
+# Positional equivalent of `GNNGraph(g, graph_type = :coo)`. Must stay kwarg-free:
+# GNNGraphsEnzymeCoreExt marks it EnzymeRules.inactive, and the rule only covers
+# the body of a positional method.
```
wsmoses: enzymerules support kwargs so you don't need to do this?

GNNGraphs/test/ext/EnzymeCore.jl
```diff
@@ -0,0 +1,59 @@
...
+        @test get_graph_type(gcoo) == :coo
+        @test gcoo.num_nodes == gref.num_nodes == g.num_nodes
+        @test gcoo.num_edges == gref.num_edges == g.num_edges
+        @test edge_index(gcoo) == edge_index(gref)
+        @test get_edge_weight(gcoo) == get_edge_weight(gref)
+        @test is_coalesced(gcoo) == false
+        @test gcoo.ndata === g.ndata
+        @test gcoo.edata === g.edata
+        @test gcoo.gdata === g.gdata
+    end
```
CarloLucibello: this has nothing to do with enzyme. these tests, if needed, belong to test/convert

GNNGraphs/test/ext/EnzymeCore.jl
```diff
@@ -0,0 +1,59 @@
...
+    # Zygote keeps differentiating through the helper (no ChainRules rule on it).
+    loss(x, g) = sum(abs2, NNlib.gather(x, edge_index(GNNGraphs._to_coo_graph(g))[1]))
+    for graph_type in (:dense, :sparse)
+        g = rand_graph(rng, 6, 10; graph_type)
+        x = randn(rng, Float32, 3, g.num_nodes)
+        gz = gradient(x -> loss(x, g), x)[1]
+        gfd = ngradient(x -> loss(x, g), x)[1]
+        @test gz ≈ gfd rtol = 1e-4
+    end
+end
```
CarloLucibello: this is testing zygote, not enzyme, these tests do not belong here.

GNNGraphs/test/ext/EnzymeCore.jl
```diff
@@ -0,0 +1,59 @@
...
+    using .GraphsTestModule
+    using EnzymeCore: EnzymeRules
+    using Enzyme: Enzyme, Const
+    import NNlib
+
+    # Loading EnzymeCore triggers the extension, which registers the rules.
+    @test hasmethod(EnzymeRules.inactive,
+                    Tuple{typeof(GNNGraphs._findnz_idx), Matrix{Int}})
+    @test hasmethod(EnzymeRules.inactive,
+                    Tuple{typeof(GNNGraphs._sparse_structure), SparseMatrixCSC{Int, Int}})
```
CarloLucibello: these tests are not needed. we should not test the internals, which may change, we should test functionality.

GNNlib/src/layers/conv.jl
```diff
@@ -460,7 +460,7 @@ end
 # when we also have edge_weight we need to convert the graph to COO
 function sgc_conv(l, g::GNNGraph{<:ADJMAT_T}, x::AbstractMatrix,
                      edge_weight::AbstractVector)
-    g = GNNGraph(g; graph_type=:coo)
+    g = _to_coo_graph(g)
```
CarloLucibello, line 463: it is a pity we have to define an extra method that does the same job as the constructor. can't we make the constructor also differentiable?

GNNGraphs/src/convert.jl
```diff
@@ -116,6 +116,40 @@ function to_coo(adj_list::ADJLIST_T; dir = :out, num_nodes = nothing, weighted =
     (s, t, nothing), num_nodes, num_edges
 end
 
+function _sparse_structure(A::AbstractSparseMatrix)
+    s, t, _ = findnz(A)
+    return s, t
+end
```
CarloLucibello, line 125: why the need to define this method? can't we just use `s, t, _ = findnz(A)` instead? If Enzyme has problems in differentiating `findnz` it should be reported and fixed upstream

GNNlib/src/layers/conv.jl
Parvm1102: I tried this but Enzyme's throws IllegalTypeAnalysisException on union-typed keyword bodies. Even the no-op GNNGraph(g, graph_type = :coo) on a :coo graph fails, and a helper delegating to to_coo(g.graph; num_nodes = ...) fails the same way inside a full GCNConv pullback. So the helper has to be positional and keyword-free.

GNNGraphs/src/convert.jl
Parvm1102: I've removed this in the new commit.

Parvm1102:
@CarloLucibello please review this, I've updated the pr

CarloLucibello (approved):

Issue #702, opened 2026-07-30
Mooncake on CUDA issue tracker


Parvm1102:
I ran Mooncake on GPU where the test sites for CPU were enabled (Mooncake 0.5.41, CUDA.jl 6.2.1, Flux 0.16.10, Julia 1.12.5). 
CPU is fully green, but many fail on GPU. Here are my findings:

**Already works on GPU** (COO storage): GraphConv, SAGEConv, GINConv, EdgeConv, NNConv, ResGatedGraphConv, CGConv, MEGNetConv, GlobalPool, GlobalAttentionPool, GCNConv with `add_self_loops=false`, and all the `propagate` fast paths (`copy_xj` with +/mean, `e_mul_xj`, `w_mul_xj`). The rules in `GNNlibMooncakeExt`.

**Zygote uses ChainRules** for non-differentiable functions, Mooncake doesn't, so it the hits GPU code it can't differentiate:

- `add_self_loops` for COO does `nodes = convert(typeof(s), [1:n;])`, i.e. builds a CPU vector and copies it to the GPU. That lands on a CPU→GPU `unsafe_copyto!` containing try/catch, which Mooncake can't trace (its cross-device copy rule only covers float arrays, and here the destination is an integer index array). It affects all layers with self loops: GCNConv (default), GATConv, GATv2Conv, TransformerConv, AGNNConv, SGConv, TAGConv, TGCN.
- On `:dense` graphs, `edge_index`/`to_coo` call `_findnz_idx`, i.e. `findall` on a `CuArray{Bool}` (no Mooncake rule), and right after that `v = A[nz]` indexes with a vector of `CartesianIndex` (also no rule). This kills every layer on dense storage.

We can add `GNNGraphs/ext/GNNGraphsMooncakeExt.jl` (Mooncake as weakdep) giving Mooncake the same semantics the ChainRules markers give Zygote:

```julia
Mooncake.@zero_derivative Mooncake.DefaultCtx Tuple{typeof(add_self_loops), GNNGraph}
Mooncake.@zero_derivative Mooncake.DefaultCtx Tuple{typeof(GNNGraphs._findnz_idx), Any}
Mooncake.@zero_derivative Mooncake.DefaultCtx Tuple{typeof(to_coo), GNNGraphs.ADJMAT_T}
[...]

PR #704, opened 2026-08-11
Add GNNGraphsMooncakeExt.jl

Parvm1102:
Part of #702 (Mooncake on CUDA tracker).

Mooncake does not use the ChainRules `@non_differentiable` annotations that keep Zygote working, so it traces into graph-structure functions and fails on CUDA (CPU to GPU index copies, `findall` on `CuArray`, bounds checks of integer indexing). This PR gives Mooncake the equivalent rules.

### Changes

- New extension `GNNGraphs/ext/GNNGraphsMooncakeExt.jl` with three `Mooncake.@zero_derivative` rules, all on integer structure extraction:
  - `add_self_loops(::GNNGraph)`, mirroring the existing `@non_differentiable` in transform.jl
  - `_findnz_idx(::Any)`, mirroring the existing `@non_differentiable` in convert.jl
  - `_edge_values(::AbstractMatrix{<:Integer}, ...)`, integer edge values have no derivative
- Rewrote the edge weight extraction in `to_coo` and `_to_coo_graph`. A new helper `_edge_values(A, s, t) = vec(A)[s .+ (t .- 1) .* size(A, 1)]` replaces `v = A[nz]`. Mooncake has a CUDA rule for linear indexing but not for CartesianIndex indexing. With this, both conversions are ordinary traceable code and float edge weights stay differentiable under Mooncake, same as Zygote and Enzyme.
- Tests in `GNNGraphs/test/ext/Mooncake.jl`: Mooncake vs Zygote vs finite differences through `add_self_loops`, `edge_index` on dense and sparse graphs, `_to_coo_graph`, and a float-weighted dense adjacency. Mooncake parts only run on Julia >= 1.12, as in the other suites.
- `Mooncake` weakdep with compat `"0.5.24"`, version bump to `1.6.0-DEV`.

This is simpler than the sketch in #702. The original 4 rules marked all of `to_coo` zero-derivative, which silently dropped edge weight gradients for float dense adjacencies. With the rewrite those rules are not needed and that problem is gone. I checked each remaining rule is really required by probing the bare functions on GPU with the extension disabled.

### Tests

CPU (Julia 1.12.5, Mooncake 0.5.45, Enzyme 0.13.198): new testitem 12/12, full GNNGraphs suite 1346 pass and 0 fail (includes the #703 Enzyme test on the rewritten helper), GraphNeuralNetworks GCNConv/GraphConv/SGConv/TAGConv/EdgeConv items 316/316, GNNlib propagate 40/40.

GPU (RTX 3050, CUDA.jl 6.2.1), all of these errored before and now pass, gradients compared against Zygote:

| site | max diff vs Zygote |
|---|---|
| GCNConv coo with self loops, wrt x and params | 0.0 |
| GCNConv dense | 0.0 |
| SGConv coo, wrt x and params | 0.0 |
| EdgeConv, NNConv, CGConv, MEGNetConv, ResGatedGraphConv on dense | <= 1.5e-8 |
| EdgeConv on float-weighted dense | 0.0 |
| GNNlib `e_mul_xj` (matrix e) on dense | 0.0 |
| GNNlib `w_mul_xj` on dense | matches the CPU reference exactly, on a case where Zygote GPU is wrong (the `broken = true` test in msgpass.jl) |

Still failing on GPU, all upstream Mooncake gaps tracked in #702: GAT, GATv2, Transformer, GMMConv, unweighted dense SGConv (missing `sum(x; dims)` rule), TAGConv and Transformer (`CuMatrix + CuMatrix` has no rule), TGCN family (`repeat` has no rule).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ Entries link to the pull request that introduced them.
 ## GNNGraphs.jl — Unreleased (towards 1.6.0)
 
 **Added**
-- Added a package extension `GNNGraphsMooncakeExt` that marks the integer graph-structure extraction (`add_self_loops`, `_findnz_idx`) as zero-derivative for Mooncake, mirroring the existing ChainRules `@non_differentiable` annotations. Together with an AD-friendly rewrite of the edge-weight extraction in `to_coo`/`_to_coo_graph` (linear instead of CartesianIndex indexing, keeping edge weights differentiable), Mooncake can now differentiate on CUDA through layers using self-loops and through adjacency-matrix graphs ([#704]).
+- Added a `Mooncake` package extension, so that Mooncake can differentiate on CUDA through `add_self_loops` and through adjacency-matrix graphs. Float edge weights of adjacency-matrix graphs stay differentiable ([#704]).
 
 **Fixed**
 - Enzyme can now differentiate through the adjacency-matrix → COO graph conversion: a new internal keyword-free helper `_to_coo_graph` avoids the union-typed keyword handling of the `GNNGraph(g; graph_type)` constructor and of `to_coo`, which Enzyme's type analysis cannot compile. Together with upstream fixes in Enzyme ≥ 0.13.197 this removes the `EnzymeInternalError` crash on `:dense`/`:sparse` graphs ([#703]).
@@ -232,6 +232,6 @@ Lux implementations of the graph convolutional, pooling, and temporal layers
 [#623]: https://github.com/JuliaGraphs/GraphNeuralNetworks.jl/issues/623
 [#695]: https://github.com/JuliaGraphs/GraphNeuralNetworks.jl/pull/695
 [#696]: https://github.com/JuliaGraphs/GraphNeuralNetworks.jl/pull/696
-[#704]: https://github.com/JuliaGraphs/GraphNeuralNetworks.jl/pull/704
 [#703]: https://github.com/JuliaGraphs/GraphNeuralNetworks.jl/pull/703
+[#704]: https://github.com/JuliaGraphs/GraphNeuralNetworks.jl/pull/704
 [FluxML/Zygote.jl#1662]: https://github.com/FluxML/Zygote.jl/issues/1662
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and the packages adhere to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 Entries link to the pull request that introduced them.
 
-## GNNGraphs.jl — Unreleased (towards 1.5.2)
+## GNNGraphs.jl — Unreleased (towards 1.6.0)
+
+**Added**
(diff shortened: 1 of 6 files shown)


Parvm1102:
@CarloLucibello , please review this pr

GNNGraphs/src/convert.jl
```diff
@@ -80,9 +80,15 @@ end
 
 CRC.@non_differentiable _findnz_idx(A)
 
+# Values of `A` at the extracted edge positions, via AD-friendly linear indexing
+# (works on GPU, unlike CartesianIndex indexing). The integer-adjacency method is
+# non-differentiable (Mooncake rule in GNNGraphsMooncakeExt); float values stay
+# on the AD path.
+_edge_values(A::AbstractMatrix, s, t) = vec(A)[s .+ (t .- 1) .* size(A, 1)]
```
CarloLucibello: why the integer case was made with zero derivatives?
Does also zygote return zero derivatives for this case?

GNNGraphs/test/ext/Mooncake.jl
```diff
@@ -0,0 +1,63 @@
...
+            @test gm≈gfd rtol=1e-4
+        end
+
+        # Gradients through `_to_coo_graph`, used by the GCNConv/SGConv/TAGConv
+        # adjacency-matrix fallbacks in GNNlib.
+        for graph_type in (:dense, :sparse)
+            g = rand_graph(rng, 6, 10; graph_type)
+            x = randn(rng, Float32, 3, g.num_nodes)
+            loss(x) = sum(abs2,
+                NNlib.gather(x, edge_index(GNNGraphs._to_coo_graph(g))[1]))
```
CarloLucibello: we shouldb't be testing internal functions like `GNNGraphs._to_coo_graph`, but the public API

CHANGELOG.md
```diff
@@ -10,7 +10,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and the packages adhere to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 Entries link to the pull request that introduced them.
 
-## GNNGraphs.jl — Unreleased (towards 1.5.2)
+## GNNGraphs.jl — Unreleased (towards 1.6.0)
+
+**Added**
+- Added a package extension `GNNGraphsMooncakeExt` that marks the integer graph-structure extraction (`add_self_loops`, `_findnz_idx`) as zero-derivative for Mooncake, mirroring the existing ChainRules `@non_differentiable` annotations. Together with an AD-friendly rewrite of the edge-weight extraction in `to_coo`/`_to_coo_graph` (linear instead of CartesianIndex indexing, keeping edge weights differentiable), Mooncake can now differentiate on CUDA through layers using self-loops and through adjacency-matrix graphs ([#704]).
```
CarloLucibello: _findnz_idx is internal, so shouldn't be mentioned

GNNGraphs/test/ext/Mooncake.jl
Parvm1102: I have changed this and rewritten in my new commit

CHANGELOG.md
Parvm1102: now updated

GNNGraphs/src/convert.jl
Parvm1102: No, Zygote returns non zero gradients here. Array{<:Integer} is NoTangent anyway, so the rule loses nothing. It is needed because Mooncake's CUDA getindex rule covers just float arrays, so the integer path fails on GPU.

CarloLucibello: > Array{<:Integer} is NoTangent anyway,

what do you mean by this? Also, the zygote gradient for integer input is not necessarily an integer. 

I don't like zygote and mooncake derivatives giving different results, but if there are no better ways to do it then let's just do it and move on.

PR #706, opened 2026-08-21
Add GNNGraphsEnzymeCoreExt marking scaled_laplacian inactive for Enzyme

Parvm1102:
`scaled_laplacian` is already `@non_differentiable` for ChainRules, but Enzyme does not
read those declarations, so it differentiated the Krylov eigensolve inside it and failed
with an `IllegalTypeAnalysisException` in SparseArrays. The Laplacian depends only on the
graph, never on the node features, so treating it as constant loses no gradient.

This adds a small `EnzymeCore` package extension (same pattern as the Mooncake extension)
with a single `EnzymeRules.inactive` declaration, plus a test.

Unblocks `ChebConv`, `GConvGRUCell` and `GConvLSTMCell` under Enzyme; `GConvGRU`/`GConvLSTM`
additionally need a separate `GNNRecurrence` fix.


Parvm1102:
Please review this @CarloLucibello

PR #708, opened 2026-08-28
Enable Enzyme gradient tests across the test suite (Julia ≥ 1.12) 

Parvm1102:
### It is draft because NNlibv 0.9.45 bump commit is there but hasn't released yet on the julia registry. All CI checks will be failing because of that.

Another reason for being draft PR is because of enzyme, the test suite time has increased drastically. It will now take hours to complete running on CI. Most layers add 3-15 minutes of time each. TransformerConv now takes ~45 minutes,  GConvLSTM 33 min, GConvGRU 27 min, GCNConv 22 min and so on. Need to discuss how should we proceed. 

Current status for Enzyme: 
- 24 layers pass on every storage type. 
- 6 layers failing on some storage types:
  - GraphConv :dense/+ fails
  - GatedGraphConv :dense fails
  - SAGEConv :dense fails
  - GNNChain :dense fails
  - TAGConv sparse only
  - EvolveGCNOCell no-state only, by design
- 4 layers failing entirely: 
  - TGCNCell: hang
  - TGCN: hang
  - SGConv
  - DConv

The Dense failures and DConv will only require simple fixes.

### DIFF
--- a/GNNlib/test/Project.toml
+++ b/GNNlib/test/Project.toml
@@ -1,4 +1,5 @@
 [deps]
+Enzyme = "7da242da-08ed-463a-9acd-ee780be4f1d9"
 FiniteDifferences = "26cc04aa-876d-5657-8c51-4c34ba976000"
 Flux = "587475ba-b771-5e3f-ad9e-33799f191a9c"
 Functors = "d9f16b24-f501-4c13-a1f2-28368ffc5196"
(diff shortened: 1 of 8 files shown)


CarloLucibello:
NNlib should be registered in a few minutes

Parvm1102:
The GNN julia 1 job terminated because the resource limits. Locally the TransformerConv took around 7gb to run, plus other tests might have exhausted the resources

